PR #1, opened 2026-05-06
feat: add tool preflighting with amp

nibalizer:
probably don't want to merge this but this is basically enough to get amp working:

image

when atryum+amp plugin is installed/running but not working:
image

when atryum is working amp will pause on all tool calls pending atryum approval


nibalizer:
This made it in via @even-steven's pr

PR #2, opened 2026-05-11
Nibz/postgres support

nibalizer:
addds postgres support


nibalizer:
we're just using @hunner's code on this one

PR #5, opened 2026-05-13
Implement task-augmented requests

hunner:
As per https://modelcontextprotocol.io/specification/2025-11-25/basic/utilities/tasks#protocol-messages

This allows for async tool calls that avoid timeouts while waiting for approval.

The way it works is that the mcp spec protocol is negotiated then persisted via `MCP-Session-Id` as per [the spec](https://modelcontextprotocol.io/specification/2025-11-25/basic/transports#session-management) in a kv store (in memory or redis), and if 2025-11-25 is supplied then harnesses are coerced to use tasks by listing tools with `taskSupport: required`, but making it optional for older protocols.


nibalizer:
rad

PR #23, opened 2026-05-22
nibz/docker-override-docs

nibalizer:
This adds docs on how I merge backend and atryum networks using docker overrides

### DIFF
--- a/docs/shared-backend-docker-compose.md
+++ b/docs/shared-backend-docker-compose.md
@@ -1,0 +1,400 @@
+# Running Atryum With the ValidMind Backend in Docker Compose
+
+This guide explains the local Docker Compose override setup used when Atryum and the ValidMind backend need to run together on one Mac.
+
+It is written for developers who are comfortable with application code but do not live in Docker every day.
+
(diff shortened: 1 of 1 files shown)


docs/shared-backend-docker-compose.md
```diff
@@ -0,0 +1,400 @@
...
+
+This guide explains the local Docker Compose override setup used when Atryum and the ValidMind backend need to run together on one Mac.
+
+It is written for developers who are comfortable with application code but do not live in Docker every day.
+
+## Why This Exists
+
+Atryum and the backend each have their own `docker-compose` stack. If you start both stacks without coordination, you get three practical problems:
+
+- Both stacks try to publish Keycloak on `localhost:8088`.
```
hunner, line 11: I've moved the atryum keycloak to `8089`. I use keycloak for atryum auth instead of auth0, so I do still want to use it. I also automated the DCR configuration of keycloak to make agent authentication easier, but don't want to apply that stuff to the validmind keycloak.

This actually clears up a lot of the following auth0/keycloak stuff too.

PR #26, opened 2026-05-27
feat: db-backed agent sync config with Settings UI

even-steven:
# Pull Request Description

## What and why?

Moves agent sync configuration (org, record type, constitution field) from the static `atryum.toml` file into a database-backed `agent_sync_settings` table, managed through a new Settings page in the Atryum admin UI.

Previously an operator had to SSH in and edit a TOML file to change which ValidMind org or record type Atryum synced agents from. Now they can do it from the UI with live dropdowns populated from the VM API.

**Key changes:**
- **Migration 011**: new `agent_sync_settings` singleton table. `INSERT OR REPLACE` upsert tolerates the row being deleted without breaking GET.
- **`AgentSyncSettingsRepo`**: `Get()` returns empty settings (not an error) when no row exists; `Save()` upserts.
- **`OpenDatabase`**: `SetMaxOpenConns(1)` for SQLite — prevents stale reads caused by multiple pool connections each seeing their own snapshot.
- **Admin API**: `GET/PUT /api/v1/admin/settings`; proxy endpoints for VM orgs, record types, and custom fields; auto-syncs agents on save; deletes all agents when org or record type changes.
- **`backendclient`**: `FetchOrganizations`, `FetchPrimaryRecordTypes`, `FetchCustomFields`.
- **`config`**: removed `AgentSyncConfig` and `AIEvaluationConfig` — these are now DB-only.
- **`main`**: reads sync settings exclusively from DB at startup and on every sync call.
- **`atryum.example.toml`**: documents that agent sync is configured via UI.

## How to test

1. Start Atryum with no `[agent_sync]` in `atryum.toml`.
2. Open the admin UI → Settings.
3. Select an org, record type, and constitution field from the dropdowns.
4. Click **Save Settings** — agents should sync automatically.
5. Verify agents appear on the Agents page.
6. Change the org or record type — confirm the delete-agents warning dialog appears and agents are deleted before re-syncing.
7. Delete the `agent_sync_settings` row directly via sqlite3 and reload Settings — should load without error (empty dropdowns).

## What needs special review?

- `SetMaxOpenConns(1)` on the SQLite connection pool — this serialises all DB access. Safe for Atryum's single-instance deployment, but worth noting.
- The `INSERT OR REPLACE` upsert pattern for the singleton settings row.

- **Breaking**: removes TOML `[agent_sync]` and `[ai_evaluation]` fields. Operators must configure sync via the UI after upgrading.
- Depends on the VM backend PR (internal discovery endpoints) being deployed first.
- No env var changes.

## Release notes

Atryum agent sync configuration (org, record type, constitution field) is now managed through the Settings page in the admin UI instead of the `atryum.toml` file. Settings are persisted in the database and take effect immediately, including an automatic agent re-sync on save.

## Checklist
- [x] What and why
- [ ] Screenshots or videos (Frontend)
- [x] How to test
- [x] Labels applied
- [ ] PR linked to Shortcut
- [x] Tested locally

### DIFF
--- a/internal/store/agent_sync_settings.go
+++ b/internal/store/agent_sync_settings.go
@@ -1,0 +1,103 @@
+package store
+
+import (
+	"context"
+	"database/sql"
+	"fmt"
+	"log"
+	"time"
+
+	sq "github.com/Masterminds/squirrel"
+)
+
+// AgentSyncSettings holds the three configurable values that control agent
+// synchronisation from the ValidMind backend. All fields are empty strings
+// when the user has not yet configured sync (i.e. no row exists yet).
+type AgentSyncSettings struct {
+	OrgCUID             string
+	AgentRecordTypeSlug string
+	ConstitutionFieldKey string
+	UpdatedAt           time.Time
+}
+
+var agentSyncSettingsColumns = []string{
+	"org_cuid", "agent_record_type_slug", "constitution_field_key", "updated_at",
+}
+
+// AgentSyncSettingsRepo provides read/write access to the singleton
+// agent_sync_settings row.
+type AgentSyncSettingsRepo struct {
+	db      *sql.DB
+	sb      sq.StatementBuilderType
+	dialect Dialect
+}
+
+func NewAgentSyncSettingsRepo(db *sql.DB) *AgentSyncSettingsRepo {
+	return NewAgentSyncSettingsRepoWithDialect(db, DialectSQLite)
+}
+
+func NewAgentSyncSettingsRepoWithDialect(db *sql.DB, dialect Dialect) *AgentSyncSettingsRepo {
+	return &AgentSyncSettingsRepo{db: db, sb: statementBuilderForDialect(dialect), dialect: dialect}
+}
+
+// Get returns the current agent sync settings. Returns an empty
+// AgentSyncSettings (no error) when no row has been saved yet.
+func (r *AgentSyncSettingsRepo) Get(ctx context.Context) (AgentSyncSettings, error) {
+	query, args, err := r.sb.Select(agentSyncSettingsColumns...).
+		From("agent_sync_settings").
+		Where(sq.Eq{"id": 1}).
+		ToSql()
+	if err != nil {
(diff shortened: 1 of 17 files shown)


internal/store/agent_sync_settings.go
```diff
@@ -0,0 +1,103 @@
...
+	query, args, err := r.sb.Select(agentSyncSettingsColumns...).
+		From("agent_sync_settings").
+		Where(sq.Eq{"id": 1}).
+		ToSql()
+	if err != nil {
+		return AgentSyncSettings{}, fmt.Errorf("build agent_sync_settings select: %w", err)
+	}
+	s, err := scanAgentSyncSettings(r.db.QueryRowContext(ctx, query, args...))
+	if err == sql.ErrNoRows {
+		return AgentSyncSettings{}, nil
```
mdeyell-valid-mind, line 57: i think we should log something here

internal/store/agent_sync_settings.go
```diff
@@ -0,0 +1,103 @@
...
+
+// Get returns the current agent sync settings. Returns an empty
+// AgentSyncSettings (no error) when no row has been saved yet.
+func (r *AgentSyncSettingsRepo) Get(ctx context.Context) (AgentSyncSettings, error) {
+	query, args, err := r.sb.Select(agentSyncSettingsColumns...).
+		From("agent_sync_settings").
+		Where(sq.Eq{"id": 1}).
+		ToSql()
+	if err != nil {
+		return AgentSyncSettings{}, fmt.Errorf("build agent_sync_settings select: %w", err)
```
mdeyell-valid-mind, line 52: i think we should log the error here since the callers just ignore it

internal/store/agent_sync_settings.go
even-steven: Done — added a log on the ErrNoRows path so it\'s clear when the system is unconfigured, and a log on any scan error before returning.

internal/store/agent_sync_settings.go
even-steven: Done — added log.Printf before both error returns in Get() so build and scan failures are visible in process logs even when callers discard the error.

PR #28, opened 2026-05-28
Store LLM invocation summaries

mdeyell-valid-mind:
## Summary
We want to be able to provide llm summarizations of invocations

- add an admin summarize endpoint that sends invocation details to the backend LLM summarization endpoint and persists the returned summary
- store invocation summaries in the database and expose them through get/list responses
- add API, backend client, service, and store coverage for invocation summaries
- include summaries in SSE/list dedupe signatures so summary-only updates refresh the live UI

## Tests
- GOCACHE=/tmp/atryum-go-build go test ./internal/api ./internal/backend ./internal/invocation ./internal/store
- git diff --check

Checkout https://github.com/validmind/backend/pull/3182 and https://github.com/validmind/frontend/pull/2593
Go to the invocations tab and summarize results.


hunner (approved):
🐫

Probably don't need two migrations in one PR, but also meh.

PR #31, opened 2026-05-29
Reuse recent explicit approvals for duplicate requests

mdeyell-valid-mind:
## Summary
- add an approval reuse window for authenticated agents so duplicate tool calls can proceed without a second manual approval
- limit reuse to explicit human approvals for the same agent, server/source, tool, and serialized request arguments
- include recently succeeded invocations in the reusable approval lookup, in addition to approved/executing invocations
- add SQLite/Postgres predicate coverage and invocation tests for same-args reuse, changed-args rejection, and cross-agent isolation

## Testing
- go test ./internal/invocation ./internal/store ./cmd/atryum

I manually tested this by setting all rules to human approve.

Then i ran the curl twice. The first time I required human approval. The second time it automatically approved via rule.
```
curl -sS localhost:8080/mcp/shortcut \
    -H "Authorization: Bearer $TOKEN" \
    -H 'Content-Type: application/json' \
    -d '{"jsonrpc":"2.0","id":1,"method":"tools/call","params":{"name":"teams-list","arguments":{"x":1}}}'
```

Changing the arguments led to human approval and not another auto approval so atryum does check the tool call.

### DIFF
--- a/internal/invocation/service.go
+++ b/internal/invocation/service.go
@@ -137,7 +137,7 @@ type invocationRepo interface {
 	UpdateSummary(ctx context.Context, id string, summary string) error
 	Get(ctx context.Context, id string) (Invocation, error)
 	GetByIdempotencyKey(ctx context.Context, key string) (Invocation, error)
-	FindRecentExplicitApproval(ctx context.Context, agentID string, upstream string, tool string, since time.Time) (Invocation, error)
+	FindRecentExplicitApproval(ctx context.Context, agentID string, upstream string, tool string, input []byte, since time.Time) (Invocation, error)
 	List(ctx context.Context, filter InvocationListFilter) ([]Invocation, int, error)
 	ListAgentIDs(ctx context.Context) ([]string, error)
 }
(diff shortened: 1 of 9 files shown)


internal/invocation/service.go
```diff
@@ -252,6 +261,31 @@ func (s *Service) Invoke(ctx context.Context, req CreateInvocationRequest) (Invo
 	if err := s.invocations.Create(ctx, inv); err != nil {
 		return InvocationResponse{}, err
 	}
+	if agentID != "" {
+		if approved, err := s.invocations.FindRecentExplicitApproval(ctx, agentID, upstream.Name, req.Tool, inv.Input, now.Add(-s.approvalReuseWindow)); err == nil {
```
hunner, line 265: This will work for multiple calls of the same request, but the double-request of
1. pre-tool hook approval for mcp tool call
2. actual mcp tool call

ends up with not matching upstreams, tools, or inputs and still needs to be approved twice:

Image

hunner (commented):
.

mdeyell-valid-mind:
I moved this back to draft since it's not actively being developed

PR #34, opened 2026-06-02
[SC-16493] Add /models/<vm_cuid>/invocations reporting endpoint

even-steven:
# Pull Request Description

## What and why?

Adds a new API-key-protected endpoint `GET /models/<vm_cuid>/invocations` that allows the ValidMind backend to retrieve paginated invocation logs for a given ValidMind inventory model CUID.

The endpoint resolves the `vm_cuid` to the associated Atryum `agent_ids` via the agents store, then queries invocations with `WHERE agent_id IN (...)` across all linked runtime agents. This bridges the ValidMind model identity space to Atryum's runtime agent identity space without requiring a schema migration on the invocations table.

**Before:** No way to query invocations by ValidMind model identity.  
**After:** The ValidMind backend (and any API-key holder) can retrieve invocations for a model by CUID.

## How to test

1. Ensure an agent exists in Atryum with a `vm_cuid` set.
2. Call the endpoint with a valid API key:
   ```
   curl -H "X-API-Key: <key>" "http://localhost:8080/models/<vm_cuid>/invocations?offset=0&limit=20"
   ```
3. Verify a paginated response containing invocations for the linked agent(s) is returned.
4. With an unknown `vm_cuid`, verify a 404 is returned.
5. With a `vm_cuid` that has no linked agent IDs, verify an empty invocations list is returned.

## What needs special review?

- `invocationsByVMCUID` handler path parsing — the `vm_cuid` is extracted from the URL segment after `/models/` and before `/invocations`. Confirm the trimming logic handles edge cases.
- `applyInvocationFilter` in `sqlite.go` — the `AgentIDs` `IN` clause uses `sq.Eq{"agent_id": []string{...}}`, which squirrel translates to `IN (...)`. Verify this is correct for 0-length and 1-length slices.

- No database schema changes. The `vm_cuid` lookup is done at query time via the existing `agents` table.
- No breaking changes to existing endpoints.
- Requires the calling service (ValidMind backend) to have a valid Atryum API key configured.

## Release notes

Internal infrastructure change; no user-facing release notes required.

## Checklist
- [x] What and why
- [ ] Screenshots or videos (Frontend)
- [x] How to test
- [ ] Labels applied
- [x] PR linked to Shortcut
- [ ] Unit tests added (Backend)
- [x] Tested locally
- [ ] Documentation updated (if required)
- [ ] Environment variable additions/changes documented (if required)

### DIFF
--- a/internal/invocation/model.go
+++ b/internal/invocation/model.go
@@ -30,15 +30,15 @@ type Approval struct {
 }
 
 type Invocation struct {
-	InvocationID   string     `json:"invocation_id"`
-	RequestID      *string    `json:"request_id,omitempty"`
-	IdempotencyKey *string    `json:"idempotency_key,omitempty"`
-	Tool           string     `json:"tool"`
-	Upstream       string     `json:"upstream"`
-	Status         Status     `json:"status"`
-	Approval       *Approval  `json:"approval"`
-	MatchedRuleID  *string    `json:"matched_rule_id,omitempty"`
-	AgentID        *string    `json:"agent_id,omitempty"`
+	InvocationID   string    `json:"invocation_id"`
+	RequestID      *string   `json:"request_id,omitempty"`
+	IdempotencyKey *string   `json:"idempotency_key,omitempty"`
+	Tool           string    `json:"tool"`
+	Upstream       string    `json:"upstream"`
+	Status         Status    `json:"status"`
+	Approval       *Approval `json:"approval"`
+	MatchedRuleID  *string   `json:"matched_rule_id,omitempty"`
+	AgentID        *string   `json:"agent_id,omitempty"`
 	// ClientName / ClientVersion mirror the MCP `initialize.clientInfo`
 	// the harness reported on the connection that issued this invocation.
 	// Captured even when auth is disabled (no agent_id) so the UI can still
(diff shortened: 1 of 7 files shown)


mdeyell-valid-mind:
nit can you add a unit test

internal/invocation/model.go
```diff
@@ -68,6 +68,7 @@ type InvocationListFilter struct {
 	Tool       string
 	Status     string
 	AgentID    string
+	AgentIDs   []string // when set, filters to invocations whose agent_id is in this list
```
mdeyell-valid-mind: i think removing the old AgentID and using AgentIDS everywhere would be nice

hunner (approved):
🕵️‍♀️

nibalizer (approved):

internal/invocation/model.go
even-steven: Done — removed `AgentID string` from `InvocationListFilter` entirely. `invocationsByAgentID` now sets `AgentIDs: []string{agentID}`, and the store collapses to a single `if len(filter.AgentIDs) > 0` branch. All existing tests updated too.

even-steven:
Added `TestInvocationsByVMCUID` in `internal/api/handlers_test.go` with four sub-tests: happy path (known vm_cuid returns invocations), 404 (unknown vm_cuid), empty list (agent exists but has no agent_ids yet), and 400 (vm_cuid missing from path).

PR #35, opened 2026-06-04
[codex] Support no-config ValidMind API credential startup

hunner:
## Summary

- Support ValidMind user API credentials in Atryum while still preferring machine-user credentials when present.
- Allow startup without an `atryum.toml` by using code defaults, including the prod ValidMind base URL.
- Rename ValidMind-related env vars to `VM_*` and add local build recipes for quick and production-like binaries.

## Impact

`VM_API_KEY=... VM_API_SECRET=... ./atryum` can run with no config file, use local SQLite, serve the embedded UI, and connect to the prod ValidMind backend using org-bound API credentials.

## Validation

- `go test ./...`
- `just build && ./atryum -h`
- `just build-prod && ./atryum -h`

### DIFF
--- a/Justfile
+++ b/Justfile
@@ -1,6 +1,7 @@
 set shell := ["bash", "-cu"]
 
 config := "./atryum.toml"
+frontend_dir := "../frontend"
 
 # List justfile targets
 default:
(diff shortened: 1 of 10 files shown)


Justfile
```diff
@@ -34,6 +35,35 @@ fmt:
...
+	  rm -rf "$tmp_dir"
+	}
+	trap cleanup EXIT
+
+	(cd "$frontend_dir" && npm run build:atryum)
+	mkdir -p "$tmp_dir/atryum"
+	rsync -a --delete \
+	  --exclude .git \
+	  --exclude /atryum \
+	  --exclude /atryum.db \
```
hunner, line 59: Ugh, greatly dislike this as it can easily pollute the binaries with things like `.DS_Store` or whatever.

nibalizer (approved):

PR #39, opened 2026-06-04
External invocations: self-declared agent_id + demo-friendly Keycloak…

nibalizer:
… lifespans

ExternalSubmitRequest now accepts an optional `agent_id` field. When the external route runs anonymously (no OAuth identity in context — the default for the amp-plugin and shared-agent-hook examples), Submit promotes the self-declared value onto the invocation row so the Agent column populates and the agents.agent_ids -> Agent Record resolution / agent-scoped approval rules light up the same way they do for verified MCP traffic. A verified OAuth identity in context still wins, so this is purely additive.

- internal/invocation/model.go: AgentID string field on ExternalSubmitRequest
- internal/invocation/service.go: Submit falls back to req.AgentID when auth.AgentIDFromContext(ctx) is empty
- internal/invocation/agent_id_test.go: TestExternalSubmitUsesSelfDeclaredAgentIDWhenNoAuthIdentity covers both the fallback and the auth-wins-over-body invariant
- examples/amp-plugin/atryum.ts: reads ATRYUM_AGENT_ID and sends it as agent_id; omitted when unset (default behavior unchanged)
- examples/shared-agent-hook/atryum-hook.mjs: same treatment
- examples/amp-plugin/README.md: documents the new env var and the "add the string to an Agent Record's Agent IDs" UI workflow

Also bumps the Keycloak realm token/session lifespans in setup-realm.sh so OAuth-based MCP clients (Claude Code, opencode) don't expire mid-demo. Stock Keycloak defaults (300s access / 30min idle / 10h max) are why tokens felt "gone in 5-10 minutes". New defaults are 24h access token / 30d sessions, overridable via ACCESS_TOKEN_LIFESPAN, SSO_SESSION_*, CLIENT_SESSION_* env vars. Applied in a new Step 2b right after realm creation; re-run ./keycloak/setup-realm.sh against an existing realm to reconcile.

Amp-Thread-ID: https://ampcode.com/threads/T-019e80f8-23f2-74a8-9120-ca3aaf2cef16

### DIFF
--- a/examples/amp-plugin/README.md
+++ b/examples/amp-plugin/README.md
@@ -71,6 +71,29 @@ with Approve / Deny buttons.
 | `ATRYUM_URL`      | `http://localhost:8080`| base URL of the atryum server                               |
 | `ATRYUM_SOURCE`   | `amp`                  | label shown as the "upstream" column in the atryum UI       |
 | `ATRYUM_POLL_MS`  | `2000`                 | how often the plugin polls atryum while awaiting approval   |
+| `ATRYUM_AGENT_ID` | _(empty)_              | self-declared agent identifier; matched against Agent Record `agent_ids` (see below) |
+
+## Tagging invocations to an Agent Record
+
+By default the plugin sends no agent identity, so invocations show up in
+the UI unattached to any Agent Record. To get the agent column populated
+and to make agent-scoped approval rules apply:
+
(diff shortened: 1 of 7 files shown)


mdeyell-valid-mind (approved):

hunner (approved):
🤘

PR #40, opened 2026-06-05
Allow mapping an anonymous mcp agent to an Agent Record

nibalizer:
This means you can do mcp things with atryum without the oidc process - important for quickstart demo!

Backend changes:
- Store anonymous_agent_record_cuid in agent sync settings.
- Stamp anonymous MCP and agent-rules requests with the selected record CUID when no auth identity exists.
- Resolve agent records by either runtime agent_id or record CUID so constitution and agent-scoped rules apply.
- Add store/API/auth tests for the fallback behavior.
- Include redirect Location details in backend API error statuses to make http:// base URL mistakes easier to diagnose."


hunner:
We can use the `?agent_id=whatever` from #44

PR #53, opened 2026-06-10
Fixes dockerfile

gtagle:
Update Dockerfile to use Node 22-alpine as the base image, enabling the inclusion of UI assets during the build process. This change replaces the previous alpine:3.20 image and ensures that the Go application can access the necessary web assets from the Node build stage.


nibalizer:
This was merged in #56

PR #63, opened 2026-06-12
add Apache 2 license

hunner:
Adds a top-level `LICENSE` file with the Apache License 2.0 text from https://www.apache.org/licenses/LICENSE-2.0.txt

I originally added a dual license as per https://github.com/validmind/validmind-library/blob/main/LICENSE but since the Apache-2 is already friendly to proprietary activities I didn't think it was needed.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,202 @@
+
+                                 Apache License
+                           Version 2.0, January 2004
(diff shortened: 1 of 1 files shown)


mdeyell-valid-mind (approved):
in slack spencer said to use apache2

PR #66, opened 2026-06-12
Guard against duplicate agent_ids across agents

even-steven:
## Summary

- **Backend conflict check**: `CheckAgentIDConflict` added to `AgentsRepo`; called in both `POST` and `PATCH /api/v1/admin/agents` before writing — returns `409 Conflict` with a message naming the conflicting ID and its owner agent.
- **Deterministic fallback**: `GetByAgentID` now includes `ORDER BY id` so behaviour is consistent if duplicates somehow reach the DB.
- **Intra-record dedup**: `parseAgentIDs` deduplicates within a single agent's array, preventing badge-count inflation.
- **UI validation**: `CreateAgentModal` and `EditAgentModal` check the cached agent list before submitting; on conflict the error alert lists each duplicate ID and the agent that owns it (e.g. _vmclient is already in use by "AMP"_).
- **Better error extraction**: `errorMessage` helper now prefers `err.response.data.error.message` (the API body) over the raw axios status line, so backend errors like `409` surface their human-readable text.
- **Observability**: `agentByAgentID` in `Invocations.tsx` logs a `console.warn` when a collision is detected.

## Test plan

- [ ] Create two agents; add the same agent ID to both — confirm the second save is blocked with the descriptive error in the UI
- [ ] Verify the same ID can be re-saved on the agent that already owns it (edit modal excludes self from the check)
- [ ] Confirm `go test ./internal/...` passes
- [ ] Confirm the Invocations page still resolves agent badges correctly with non-duplicate data

### DIFF
--- a/internal/api/handlers.go
+++ b/internal/api/handlers.go
@@ -2694,6 +2694,15 @@ func (h *Handler) adminAgentDetail(w http.ResponseWriter, r *http.Request) {
 			writeError(w, http.StatusBadRequest, "invalid json")
 			return
 		}
+		if req.AgentIDs != nil && len(req.AgentIDs) > 0 {
+			if conflictID, ownerName, err := h.agentsRepo.CheckAgentIDConflict(r.Context(), id, req.AgentIDs); err != nil {
+				writeError(w, http.StatusInternalServerError, "failed to check agent ID conflicts")
+				return
+			} else if conflictID != "" {
+				writeError(w, http.StatusConflict, fmt.Sprintf("agent ID %q is already claimed by agent %q", conflictID, ownerName))
+				return
+			}
+		}
 		if err := h.agentsRepo.UpdateEnabled(r.Context(), id, req.Enabled); err != nil {
 			status := http.StatusInternalServerError
 			if err == sql.ErrNoRows {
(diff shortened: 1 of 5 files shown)


internal/api/handlers.go
```diff
@@ -2684,6 +2703,15 @@ func (h *Handler) adminAgentDetail(w http.ResponseWriter, r *http.Request) {
...
 		if req.AgentIDs != nil {
+			if len(req.AgentIDs) > 0 {
+				if conflictID, ownerName, err := h.agentsRepo.CheckAgentIDConflict(r.Context(), id, req.AgentIDs); err != nil {
+					writeError(w, http.StatusInternalServerError, "failed to check agent ID conflicts")
+					return
+				} else if conflictID != "" {
+					writeError(w, http.StatusConflict, fmt.Sprintf("agent ID %q is already claimed by agent %q", conflictID, ownerName))
+					return
+				}
+			}
```
mdeyell-valid-mind: We should check this before calling `UpdateEnabled`

even-steven: Done

mdeyell-valid-mind (approved):

PR #67, opened 2026-06-12
[SC-16740] Remove invocation ID column from table

even-steven:
## Summary

- Removes the **ID** column (`invocation_id`) from the invocations table — the full ID is still accessible in the detail panel when a row is selected
- Removes the now-unused `truncateMiddle` helper

## Test plan

- [ ] Open the Invocations page and confirm the ID column is gone
- [ ] Click a row and confirm the invocation ID is still visible in the detail panel


mdeyell-valid-mind (approved):
seems reasonable

PR #74, opened 2026-06-14
Pull get rules tool

nibalizer:
Before this change we had an atryum specific tool on every server called `atryum_rules_get` that would let the agent inspect static tools. 

The idea here was to help direct the agent to use tools that will auto approve when that will suffice instead of hitting a `pending_human` review that stops the work (slows the agent down to human speed).

However in practice this had some issues:

- agents were reluctant to call tools they knew would be denied, making it harder to show atryum's gate in action
- the `atryum.rules.get` tool name is spec compliant but broke on some common harness implementations (cursor, claude desktop)
- harness(invocations api) and orchestrator level agents didn't have implementations
- persistent security concerns about giving the agent more information about it's sandbox mechanism

Removing this temporarily and we will land a more complete implementation later.


hunner (approved):
🫆

PR #76, opened 2026-06-15
ci: publish atryum image to Docker Hub

eggshell:
## Summary

Adds a GitHub Actions workflow that builds the production atryum image from `Dockerfile.prod` and publishes it to Docker Hub under `validmind/atryum`.

## Changes

- New workflow `.github/workflows/docker-publish.yml`
- Multi-arch builds (`linux/amd64`, `linux/arm64`) via QEMU + Buildx
- Tags derived with `docker/metadata-action`:
  - push to `main` → `latest`, `main`, `sha-<short>`
  - `v*` tags → semver (`1.2.3`, `1.2`) + `sha-<short>`
  - manual `workflow_dispatch` → custom tag input (defaults to `latest`)
- GitHub Actions build cache to speed up rebuilds

## Notes

The org `gha-workflows` reusable build/push workflows target AWS ECR only, so a standalone workflow is used here to reach the Docker Hub destination.

## Requirements

Repo/org secrets must be set: `DOCKERHUB_USERNAME` and `DOCKERHUB_TOKEN`.

### DIFF
--- a/.github/workflows/docker-publish.yml
+++ b/.github/workflows/docker-publish.yml
@@ -1,0 +1,65 @@
+name: Publish Docker image
+
+on:
+  push:
+    branches: [main]
+    tags: ["v*"]
+  workflow_dispatch:
+    inputs:
+      tag:
+        description: "Image tag to push (defaults to 'latest')"
+        required: false
+        default: latest
+
+permissions:
+  contents: read
+
+concurrency:
+  group: docker-publish-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  build-push:
+    name: Build and push to Docker Hub
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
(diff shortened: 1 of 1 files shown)


hunner (approved):
🏗️

nibalizer (approved):

cachafla:
Merge?

hunner:
Okay I think I lost the plot on this one. We currently ship as a single binary with everything embedded, so do we still need/want to ship a docker container? Is this so we can ship it in validmind's helm chart maybe? Or the "npx run mcp servers locally" may be nice to contain in a docker container instead of directly on your host.

The [current docs](https://atryum.org/documentation/1_quickstart.html#download-atryum) mention running via the docker compose stack, so if we wanted to come up with some prescriptive scenarios about running this from a published docker container we would want to revise that.

PR #77, opened 2026-06-15
fix(pi): Add judge reject reason in pi agent hook

hunner:
The llm judge returns a reject reason, but `pi` only ever saw ` atryum: tool call 'bash' was denied by reviewer.` This exposes the reason back to the agent to adjust their behavior.


mdeyell-valid-mind (approved):
good catch

PR #95, opened 2026-06-16
fix(oauth): Refresh OAuth tokens for MCP upstreams

hunner:
Replace the non-refreshing oauth credential store with a refreshing one.


mdeyell-valid-mind (approved):
works for me with shortcut

nibalizer (approved):

PR #101, opened 2026-06-17
Always have temperature as zero when talking to the llm

mdeyell-valid-mind:
Sets temperature to 0.0 on every request the local evaluator sends to an LLM — both the tool-call judge and the invocation summarizer

Why

The judge and summarizer are deterministic decision-making tasks: a given charter + tool call should yield the same verdict every time, and summaries shouldn't vary run-to-run. Previously no temperature was specified, so we inherited each provider's default (typically 1.0), introducing avoidable nondeterminism into safety verdicts and audit summaries.

Some models might not like us sending a temperature parameter so we may want to over complicate this change in the future by adding a temperature column to LLMConfig table but it's probably worth waiting until we notice a model not working with temperature 0.0.

### DIFF
--- a/internal/invocation/local_evaluator.go
+++ b/internal/invocation/local_evaluator.go
@@ -217,7 +217,7 @@ func (e *LocalEvaluatorClient) callAnthropic(ctx context.Context, cfg LocalLLMCo
 }
 
 // parseVerdict extracts verdict/confidence/reason from the LLM JSON output.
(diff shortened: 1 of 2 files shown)


mdeyell-valid-mind:
https://github.com/validmind/atryum/issues/100

hunner (approved):
🥶

Issue #97, opened 2026-06-16
Make it so that agent_id can always come from the oidc token when it is provided


mdeyell-valid-mind:
Make it so that all of the endpoints used by agent harnesses can use an oauth token.

PR #102, opened 2026-06-18
Add oauth to non admin endpoints used by agent harnesses

mdeyell-valid-mind:
Now all of the endpoints used by agent harnesses have an oauth decorator. We also update the agent harnesses to make sure that the token is passed when present. 
https://github.com/validmind/atryum/issues/97


nibalizer (approved):
We will need docs on how to get the auth setup of course

mdeyell-valid-mind:
> We will need docs on how to get the auth setup of course

good catch, updated

hunner (approved):
🪙

Issue #103, opened 2026-06-18
add authentication to admin endpoints


mdeyell-valid-mind:
We should make a new admin endpoint decorator that looks for admin oauth tokens

PR #110, opened 2026-06-22
Add admin auth

mdeyell-valid-mind:
We would like the admin endpoints to be auth protected. In order to accomplish this a new admin decorator is added to the admin endpoints. Additionally, the frontend will now prompt people to sign in and will automatically refresh tokens. 

https://github.com/user-attachments/assets/3116bc4c-bba9-4c11-b86d-fb17b4b46604

This is designed to be IDP agnostic but i have only done full testing with keycloak since this repo ships with that.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -70,7 +70,6 @@ services:
       KC_ADMIN: admin
       KC_PASSWORD: admin
       REALM: atryum
-      KC_SETUP_SKIP_IF_REALM_EXISTS: "false"
     volumes:
       - ./keycloak/setup-realm.sh:/setup-realm.sh:ro
     command:
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -70,7 +70,7 @@ services:
       KC_ADMIN: admin
       KC_PASSWORD: admin
       REALM: atryum
-      KC_SETUP_SKIP_IF_REALM_EXISTS: "true"
+      KC_SETUP_SKIP_IF_REALM_EXISTS: "false"
     volumes:
       - ./keycloak/setup-realm.sh:/setup-realm.sh:ro
     command:
--- a/internal/auth/middleware.go
+++ b/internal/auth/middleware.go
@@ -81,13 +81,12 @@ func AdminMiddleware(v *Validator, opts MiddlewareOptions) func(http.Handler) ht
 		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 			header := strings.TrimSpace(r.Header.Get("Authorization"))
 			if header == "" {
-				if opts.DebugLogIdentity {
-					logAuthFailure(r, "invalid_token", "missing bearer token", "")
-				}
+				logAuthFailure(r, "invalid_token", "missing bearer token", "")
 				writeChallenge(w, http.StatusUnauthorized, "missing bearer token", "", "", "")
 				return
 			}
 			if !strings.HasPrefix(strings.ToLower(header), "bearer ") {
+				logAuthFailure(r, "invalid_request", "invalid Authorization scheme", "")
 				writeChallenge(w, http.StatusUnauthorized, "invalid Authorization scheme", "invalid_request", "", "")
 				return
 			}
--- a/internal/auth/middleware.go
+++ b/internal/auth/middleware.go
@@ -73,6 +73,49 @@ func MiddlewareWithOptions(v *Validator, resourceMetadataPath string, opts Middl
 	}
 }
 
+func AdminMiddleware(v *Validator, opts MiddlewareOptions) func(http.Handler) http.Handler {
+	return func(next http.Handler) http.Handler {
+		if v == nil || !v.AdminEnabled() || opts.SkipVerify {
+			return next
+		}
+		return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			header := strings.TrimSpace(r.Header.Get("Authorization"))
+			if header == "" {
+				if opts.DebugLogIdentity {
+					logAuthFailure(r, "invalid_token", "missing bearer token", "")
+				}
+				writeChallenge(w, http.StatusUnauthorized, "missing bearer token", "", "", "")
+				return
+			}
+			if !strings.HasPrefix(strings.ToLower(header), "bearer ") {
+				writeChallenge(w, http.StatusUnauthorized, "invalid Authorization scheme", "invalid_request", "", "")
+				return
+			}
+			token := strings.TrimSpace(header[len("Bearer "):])
+			identity, err := v.ValidateAdmin(r.Context(), token)
+			if err != nil {
+				ve, _ := err.(*ValidationError)
+				switch {
+				case ve != nil && ve.Result == ResultMissingAdminClaim:
+					logAuthFailure(r, "insufficient_scope", ve.Description, "")
+					writeChallenge(w, http.StatusForbidden, ve.Description, "insufficient_scope", "", "")
+				case ve != nil:
+					logAuthFailure(r, "invalid_token", ve.Description, "")
+					writeChallenge(w, http.StatusUnauthorized, ve.Description, "invalid_token", "", "")
+				default:
+					logAuthFailure(r, "invalid_token", "invalid token", "")
+					writeChallenge(w, http.StatusUnauthorized, "invalid token", "invalid_token", "", "")
+				}
+				return
+			}
+			if opts.DebugLogIdentity {
+				log.Printf("[auth] valid_admin_token method=%s path=%s remote=%s issuer=%q subject=%q email=%q", r.Method, r.URL.Path, r.RemoteAddr, identity.Issuer, identity.Subject, identity.Email)
+			}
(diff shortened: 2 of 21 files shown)


docker-compose.yml
```diff
@@ -70,7 +70,7 @@ services:
       KC_ADMIN: admin
       KC_PASSWORD: admin
       REALM: atryum
-      KC_SETUP_SKIP_IF_REALM_EXISTS: "true"
+      KC_SETUP_SKIP_IF_REALM_EXISTS: "false"
```
mdeyell-valid-mind: this is so that existing installs using keycloak will get the ` "atryum_admin": true,` field in their token

hunner: Everything in the keycloak setup script should be idempotent, so we probably don't need this env var any more either. Other than making startup faster.

mdeyell-valid-mind: removed

mdeyell-valid-mind:
https://github.com/validmind/atryum/issues/103

website/md-drafts/1_integrations/2_connect-agents.md
```diff
@@ -370,3 +370,45 @@ For deployments outside local development, use your organization's identity prov
 :::
 When auth mode is enabled, configure your agent to send a bearer token rather than a query-parameter agent ID. Amp, Pi, and the shared hook script read `ATRYUM_ACCESS_TOKEN` and send it to Atryum's agent runtime APIs.
 :::
+
+### Admin UI authentication
+
+Admin UI/API authentication is opt-in. If no configured `[[auth]]` block has `admin_enabled = true`, `/ui/` and the admin API remain open for local development. As soon as one or more blocks are admin-enabled, Atryum requires a valid browser OIDC access token for admin API calls. The upstream MCP OAuth callback at `/api/v1/admin/oauth/callback` remains public so external identity providers can complete browser redirects.
```
hunner, line 376: I wonder if it makes sense to mv `/api/v1/admin/oauth/callback` to `/api/v1/mcp/oauth/callback` since it's the oauth callback for upstream mcp sever connection flows and not admin related.

website/md-drafts/1_integrations/2_connect-agents.md
```diff
@@ -370,3 +370,45 @@ For deployments outside local development, use your organization's identity prov
...
+audience = "atryum"
+required_scope = "atryum:mcp"
+agent_id_claim = "client_id"
+
+admin_enabled = true
+admin_provider = "keycloak"
+admin_client_id = "atryum-admin"
+admin_scopes = "openid profile email atryum:mcp"
+admin_claim = "atryum_admin"
+admin_claim_value = true
```
hunner, line 393: This kind of config means that the admin auth and agent auth are coupled. Trying to do `enabled = false` but still have `admin_enabled = true` doesn't work. The validation just skips the whole block due to `enabled = false` and admin remains open.

Maybe it could be more clear with something like:
```toml
[[admin_auth]]
enabled = true
issuer = "http://localhost:8089/realms/atryum"
audience = "http://localhost:8080/api/v1/admin"
client_id = "atryum-admin"
scopes = "openid profile email"
claim = "atryum_admin"
claim_value = true
provider = "keycloak"
```

This uses a different `aud` for the admin auth example as well. It includes the deployment host and a url path to distinguish it from the MCP token audience (which should probably be something like `http://localhost:8080/mcp` instead of just `atryum`).

internal/auth/middleware.go
```diff
@@ -73,6 +73,48 @@ func MiddlewareWithOptions(v *Validator, resourceMetadataPath string, opts Middl
...
+				default:
+					logAuthFailure(r, "invalid_token", "invalid token", "")
+					writeChallenge(w, http.StatusUnauthorized, "invalid token", "invalid_token", "", "")
+				}
+				return
+			}
+			if opts.DebugLogIdentity {
+				log.Printf("[auth] valid_admin_token method=%s path=%s remote=%s issuer=%q subject=%q email=%q", r.Method, r.URL.Path, r.RemoteAddr, identity.Issuer, identity.Subject, identity.Email)
+			}
+			next.ServeHTTP(w, r)
```
hunner, line 113: The `AdminMiddleware` has `identity` above, but doesn't pass it into `next.ServeHTTP(w, r)` so the identity is discarded. If we send the identity along then that can become the `actor_id` value on human approvals and we get user identities that way. Eventually we could have users tracked in a table by ID, but in the mean time just grabbing the email or subject should work.

Example for `internal/auth/validator.go` on the `AdminIidentity` struct:
```go
type AdminIdentity struct {
      Issuer  string
      Subject string
      Email   string
      Name    string
}

func (id AdminIdentity) ActorID() string {
      if id.Email != "" {
              return id.Email
      }
      if id.Subject != "" && id.Issuer != "" {
              return id.Issuer + "#" + id.Subject
      }
      return id.Subject
}
```

Then in `auth/context.go`:
```go
const (
      identityKey ctxKey = iota
      adminIdentityKey
)

func WithAdminIdentity(parent context.Context, id AdminIdentity) context.Context {
      return context.WithValue(parent, adminIdentityKey, id)
}

func AdminIdentityFromContext(ctx context.Context) (AdminIdentity, bool) {
      id, ok := ctx.Value(adminIdentityKey).(AdminIdentity)
      return id, ok
}
```

Then the middleware line becomes:

```go
next.ServeHTTP(w, r.WithContext(WithAdminIdentity(r.Context(), identity)))
```

In practice, identity after ValidateAdmin might look like:

```go
auth.AdminIdentity{
      Issuer:  "https://idp.example.com/realms/company",
      Subject: "2f7d6d4e-9bb2-4c10-99cb-9cf2f93c9e4a",
      Email:   "alice@example.com",
      Name:    "Alice Admin",
}
```

Then approval handling can do:

```go
admin, _ := auth.AdminIdentityFromContext(r.Context())
actorID := admin.ActorID()
```

and thread `actorID` into `Approve` / `Deny` so `Approval.ActorID` and `DecisionAt` are populated.

internal/auth/validator.go
```diff
@@ -303,3 +353,29 @@ func stringClaim(claims jwt.MapClaims, name string) string {
...
+	got, ok := claims[claimName]
+	if !ok {
+		return false
+	}
+	want = strings.TrimSpace(want)
+	switch v := got.(type) {
+	case string:
+		return strings.TrimSpace(v) == want
+	case bool:
+		return strings.EqualFold(want, fmt.Sprintf("%t", v))
```
hunner, line 371: We should either have cases for `float64` or remove the `int64` case from `UnmarshalTOML` (toml values come in as ints, but json values come in as floats).

hunner (commented):
The docs around how to add auth kind of confused me when setting this up. I think it would be better to have a separate toml block specifically for user auth so it's very obviously distinct. This also allows user auth and agent auth to be independent.

Also we can put the user's ID on the approval records, since we actually have that now.

internal/auth/middleware.go
mdeyell-valid-mind: `The AdminMiddleware has identity above, but doesn't pass it into next.ServeHTTP(w, r) so the identity is discarded.`
This is on purpose because we don't have users in atryum, all admins are equal.

website/md-drafts/1_integrations/2_connect-agents.md
mdeyell-valid-mind: Why wouldn't we want an agent and admin to be able to use the same IDP? I think separate admin only IDP blocks add complexity without value

internal/auth/middleware.go
mdeyell-valid-mind: `Eventually we could have users tracked in a table by ID`
we specifically do not want users in atryum

website/md-drafts/1_integrations/2_connect-agents.md
hunner: Maybe I'm just used to `[[auth]]` blocks applying to agents and having multiple auth blocks is pretty common in various scenarios during development, so when I was following the docs to set up the auth I actually started setting up another block for admin auth and only noticed that I was supposed to extend an existing agent auth block afterward. I guess in most common scenarios users will only have a single `[[auth]]` block at all, so keeping admin unified makes sense.

internal/auth/middleware.go
hunner: It feels so strange to have an `Approval.ActorID` and just leave it blank, even if all users are equal. But that's fine.

website/md-drafts/1_integrations/2_connect-agents.md
mdeyell-valid-mind: good idea i will update this

internal/auth/validator.go
mdeyell-valid-mind: good catch, i'll update this

hunner (approved):
🕰️

website/md-drafts/1_integrations/2_connect-agents.md
mdeyell-valid-mind: i made a new issue for it here https://github.com/validmind/atryum/issues/120

PR #113, opened 2026-06-23
Add desktop notifications

mdeyell-valid-mind:
Whenever an agent gets blocked on human approval we would like a notification so that we know our agent is blocked.

https://github.com/user-attachments/assets/0af8d949-fc9c-4c59-980a-ce8b8bc09fdd

When testing this i first had to enable firefox/brave notifications.


hunner (approved):
🔔

Issue #91, opened 2026-06-16
Agent tool-use guidance


hunner:
Provide agents advisory responses on how to use tools within policy bounds, including which servers, tools, or argument patterns are likely to pass evaluation, without executing the call. Reduces trial-and-error loops and gives agents a lightweight compliance coach at planning time.

PR #119, opened 2026-06-23
Tell agents to read in their rules to reduce the number of denied calls

mdeyell-valid-mind:
We would like agents to be able to read in their rules to reduce the number of denied calls.
A previous version of this pr added a mcp tool for this but it was useless for the agent harnesses. This time around, we have the harnesses hit the rules endpoint every 5 minutes to get their rules. The mcp rule call also got readded.

https://github.com/validmind/atryum/issues/91

Open question:
Currently, i only expose static tools to the agent. Should we also expose llm as judge rules? I wonder if an agent knowing about llm as judge rules could make it act differently hmmmmm

Screenshot 2026-06-26 at 16 05 08

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,7 +129,7 @@ The frontend fetches `/api/v1/admin-auth/config`, shows a sign-in screen, redire
 
 Public (auth-protected when `[[auth]]` is configured):
 
-- `POST /mcp/{server}` — MCP JSON-RPC. `tools/list` annotates each tool with its policy disposition for the calling agent; a synthetic `atryum.rules.get` tool lets an agent inspect its applicable rules before deciding what to call.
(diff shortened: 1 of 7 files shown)


mdeyell-valid-mind:
https://github.com/validmind/atryum/issues/91

even-steven (commented):
Two issues worth addressing before merge: 
(1) formatRulesContext/rulesContext/combineContext are copy-pasted identically into all three example files — a shared module or at-minimum sync comments would help; 

(2) the request_id fallback in handleAtryumRulesToolCall isn't listed in the MCP tool's InputSchema.

mdeyell-valid-mind:
> (1) formatRulesContext/rulesContext/combineContext are copy-pasted identically into all three example files — a shared module or at-minimum sync comments would help;

The complication with number 1 is that these files would no longer be directly copy able ie ```  cp examples/pi-extension/index.ts /path/to/your/pi-extension/index.ts```

mdeyell-valid-mind:
> 
> (2) the request_id fallback in handleAtryumRulesToolCall isn't listed in the MCP tool's InputSchema.

I updated number 2

mdeyell-valid-mind:
rebased

even-steven (approved):
Looks good to me ✅

even-steven (approved):
👍🏽

Issue #120, opened 2026-06-26
mv /api/v1/admin/oauth/callback to /api/v1/mcp/oauth/callback


mdeyell-valid-mind:
mv /api/v1/admin/oauth/callback to /api/v1/mcp/oauth/callback since it's the oauth callback for upstream mcp sever connection flows and not admin related.

PR #121, opened 2026-06-26
mv /api/v1/admin/oauth/callback to /api/v1/mcp/oauth/callback

mdeyell-valid-mind:
the mcp oauth callback has nothing to do with the admin page so it should have a different route.

https://github.com/validmind/atryum/issues/120


hunner (approved):
⬅️

PR #122, opened 2026-06-29
[SC-16780] Allow machine keys on admin endpoints

even-steven:
## Summary

- Extends `AdminMiddleware` to accept `X-API-Key` / `X-API-Secret` as a trusted machine-caller path on admin endpoints (`/api/v1/admin/...`), in addition to the existing OAuth Bearer-token path used by the human admin UI
- When valid machine key credentials are present, the request is admitted immediately without requiring a JWT; wrong keys return 401 without falling through to Bearer
- This lets server-to-server callers (e.g. the ValidMind backend proxy) use the same static credentials they already use for reporting endpoints — no separate `admin_bearer_token` config or Keycloak service-account client-credentials flow needed

## Test plan

- [ ] Run `go test ./internal/auth/...` — all existing tests pass, new `TestAdminMiddlewareMachineKeyBypassesBearer` covers valid keys, wrong secret, and Bearer fallback
- [ ] Verify ValidMind backend can call admin endpoints using `vm_to_atryum_api_key` / `vm_to_atryum_api_secret` from vmconfig
- [ ] Verify human admin UI login via Bearer JWT still works (no `X-API-Key` headers sent from browser)

### DIFF
--- a/atryum.example.toml
+++ b/atryum.example.toml
@@ -90,10 +90,15 @@ api_key = ""          # Anthropic API key created in that workspace
 # are managed through the Atryum Settings UI and stored in the database.
 # They no longer need to be configured here.
 
-# Static API key/secret pair that protects the read-only invocation reporting
-# endpoints (GET /invocations/{agent_id}, GET /agent_ids). Callers must send
-# both `X-API-Key` and `X-API-Secret` headers matching these values. When
-# either field is empty, those endpoints refuse every request with 503.
+# Static API key/secret pair used for machine-to-machine authentication.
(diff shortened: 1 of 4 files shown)


mdeyell-valid-mind:
The code looks good but the toml comment for [api_key] should be updated. The docs should also mention that [api_key] now can be used to access the admin endpoints when it is set

even-steven:
> The code looks good but the toml comment for [api_key] should be updated. The docs should also mention that [api_key] now can be used to access the admin endpoints when it is set

Done, thank you

mdeyell-valid-mind (approved):

Issue #115, opened 2026-06-23
MCP server names with spaces break the endpoint


hunner:
Creating an MCP server with a space in its name (for example "Slack Local") produces an endpoint that clients cannot reach, and no tools are discovered. The resulting endpoint slug is not surfaced, so the failure is hard to diagnose. Success: names are validated or normalized into a usable endpoint, and the resulting URL is shown to the user.

PR #123, opened 2026-06-29
Don't allow spaces in mcp server endpoints

mdeyell-valid-mind:
Creating an MCP server with a space in its name (for example "Slack Local") produces an endpoint that clients cannot reach.

To fix this, I added a slug that can be used to select servers. Now when making a mcp server named `slack local`, it gets an internal slug of `slack-local`. The db enforces unique slugs.
I also disallow names that consist entirely of dots **.**

Screenshot 2026-06-29 at 14 40 44

- endpoint_slug storage on mcp_servers, with a migration that backfills existing servers and handles slug collisions deterministically.
  - Resolver support for looking up servers by endpoint slug as well as by original name.
  - Admin API fields for endpoint_slug and endpoint_url.
  - UI display of the resolved MCP endpoint after create/save and in the server detail panel.
  - Tests covering slug generation, duplicate slug rejection, migration backfill, resolver lookup, and admin API endpoint display.

https://github.com/validmind/atryum/issues/115


even-steven (approved):
Looks good to me 👍🏽

PR #125, opened 2026-07-02
Token harness refresh

mdeyell-valid-mind:
Adds support to the agent harness integrations so short-lived OAuth access tokens can be refreshed automatically.

I tested this with `PI`.

When reviewing please look at the docs to make sure those are usable to you.

  ## Changes

  - Added token command execution, expiry parsing, caching, and one-time 401 retry support:
  - Supports token command output as either:
    - a raw bearer token
    - JSON containing `access_token`, `accessToken`, or `token`
    - optional expiry fields via `expires_in`, `expires_at`, or `expiresAt`
  - Persists token caches for the Node integrations under each integration state directory.
  - Updates docs for auth-mode setup and short-lived token usage.


scripts/fake_agent.py
```diff
@@ -71,6 +84,29 @@
...
+    float(os.environ.get("ATRYUM_TOKEN_COMMAND_TIMEOUT_MS", "10000")) / 1000
+)
+STATE_DIR = os.environ.get("ATRYUM_STATE_DIR", "") or os.path.join(
+    os.path.expanduser("~"), ".atryum", "fake-agent-state"
+)
+TOKEN_CACHE_FILE = os.path.join(STATE_DIR, "token-cache.json") if TOKEN_COMMAND else ""
+# Ties the cached token to the command and server that produced it, so
+# switching ATRYUM_TOKEN_COMMAND or ATRYUM_URL invalidates the cache instead
+# of sending a token minted for a different identity or target.
+TOKEN_CACHE_KEY = (
```
hunner: **Issue 1: cache key uses the env URL, but requests can target `--base <url>`.** `TOKEN_CACHE_KEY` is computed from `DEFAULT_URL` at import time, while the CLI accepts `--base` (defaulting to `DEFAULT_URL`). Run once with `ATRYUM_URL` pointing at server A, then again with `--base https://server-b`, and the cached token for A is silently sent to B — the exact failure this key's comment says it prevents. Either fold the resolved base URL into the key after arg parsing, or document that `--base` bypasses the cache keying.

🤖 _Automated reply_

examples/shared-agent-hook/atryum-hook.mjs
```diff
@@ -45,14 +51,131 @@ const MAX_MESSAGE_CHARS = Number(process.env.ATRYUM_MAX_MESSAGE_CHARS || 2000);
...
+let cachedTokenExpiresAt = ACCESS_TOKEN && !TOKEN_COMMAND ? Number.POSITIVE_INFINITY : 0;
+
+function parseTokenResponse(raw) {
+  const text = String(raw || "").trim();
+  if (!text) throw new Error("token command returned no token");
+  if (!text.startsWith("{")) {
+    return { accessToken: text, expiresAt: Date.now() + 55 * 60 * 1000 };
+  }
+  const parsed = JSON.parse(text);
+  const accessToken = parsed.access_token || parsed.accessToken || parsed.token;
```
hunner: **Issue 2: this copy has weaker validation than the TS copies.** Truthy `||` with no `typeof` guard means a non-string value (e.g. a nested object) becomes `Bearer [object Object]` instead of a clear parse error. The Amp/Pi TS versions guard with `typeof === "string"` — same for the cache read below (`token &&` vs `typeof token === "string"`). Worth aligning while all four copies are still fresh.

🤖 _Automated reply_

examples/amp-plugin/atryum.ts
```diff
@@ -65,14 +73,172 @@ const CLIENT_VERSION =
...
+// switching ATRYUM_TOKEN_COMMAND or ATRYUM_URL invalidates the cache instead
+// of sending a token minted for a different identity or target.
+const TOKEN_CACHE_KEY = TOKEN_COMMAND
+  ? createHash("sha256").update(`${TOKEN_COMMAND}\n${API}`).digest("hex")
+  : "";
+let cachedToken = ACCESS_TOKEN;
+let cachedTokenExpiresAt = ACCESS_TOKEN && !TOKEN_COMMAND ? Number.POSITIVE_INFINITY : 0;
+let refreshPromise: Promise<string> | null = null;
+
+function parseTokenResponse(raw: string): {
```
hunner, line 105: **Issue 3: four hand-rolled copies, already drifting.** This file uses inline nested ternaries, the Pi extension uses its `firstString` helper, the shared hook uses truthy `||`, and the hook checks the file cache inside `accessToken()` while the TS versions check it inside `refreshAccessToken()`. I understand the examples are deliberately standalone single files, so extraction may not be desirable — but at minimum making the four `parseTokenResponse` bodies textually identical would keep future diffs reviewable.

🤖 _Automated reply_

examples/pi-extension/index.ts
```diff
@@ -8,20 +16,175 @@ const CLIENT_NAME = process.env.ATRYUM_CLIENT_NAME || SOURCE;
...
+let cachedTokenExpiresAt = ACCESS_TOKEN && !TOKEN_COMMAND ? Number.POSITIVE_INFINITY : 0;
+let refreshPromise: Promise<string> | null = null;
+
+function parseTokenResponse(raw: string): {
+  accessToken: string;
+  expiresAt: number;
+} {
+  const text = raw.trim();
+  if (!text) throw new Error("token command returned no token");
+  if (!text.startsWith("{")) {
```
hunner, line 60: **Issue 4: non-JSON stdout is trusted as a token for 55 minutes.** Any command that exits 0 and prints something that isn't JSON (a warning, an HTML error page from a proxy) gets cached and sent as a bearer token. The 401 retry saves correctness, but since the re-mint returns the same garbage, every subsequent request runs the token command plus two failing HTTP calls — a quiet thundering-herd on the IdP. A cheap sanity check (reject output containing whitespace/newlines after trim) would fail fast instead — and it's needed anyway, since an interior newline makes the header value invalid and `fetch` will throw.

🤖 _Automated reply_

integrations/lib/harness.sh
```diff
@@ -314,7 +328,9 @@ PY
   export GROK_TEST_HOME="$HARNESS_CONFIG_DIR/grok-home"
   export AMP_SETTINGS_FILE="$HARNESS_CONFIG_DIR/amp/settings.json"
   unset CLAUDE_MCP_CONFIG CLAUDE_CONFIG_DIR CURSOR_MCP_CONFIG CURSOR_RUN_DIR || true
+  unset PI_EXTENSION_PATH || true
 
+  harness_token_command_env "$auth_id"
```
hunner: **Issue 5: failure here aborts the whole run under `set -euo pipefail`.** This call is unguarded, so `static-bearer` with `ATRYUM_STATIC_BEARER_TOKEN` unset kills the entire test script with no message rather than failing that one case. Existing callers of `fetch_bearer_token` use `|| return 1`; a `|| { warn "..."; return 1; }` here would match and say why.

🤖 _Automated reply_

examples/amp-plugin/atryum.ts
```diff
@@ -65,14 +73,172 @@ const CLIENT_VERSION =
...
+  if (useFileCache) {
+    const fileCache = await readTokenCache();
+    if (fileCache) {
+      cachedToken = fileCache.token;
+      cachedTokenExpiresAt = fileCache.expiresAt;
+      return cachedToken;
+    }
+  }
+  const { stdout } = await execAsync(TOKEN_COMMAND, {
+    timeout: Number(process.env.ATRYUM_TOKEN_COMMAND_TIMEOUT_MS || 10000),
```
hunner: Nit: `ATRYUM_TOKEN_COMMAND_TIMEOUT_MS` is read inline here while every sibling env var is a hoisted const — trivial inconsistency, but it also means the README tables document a "default 10000" const that doesn't exist as one.

🤖 _Automated reply_

examples/amp-plugin/atryum.ts
```diff
@@ -65,14 +73,172 @@ const CLIENT_VERSION =
...
+const TOKEN_CACHE_FILE = TOKEN_COMMAND
+  ? join(STATE_DIR, "token-cache.json")
+  : "";
+// Ties the cached token to the command and server that produced it, so
+// switching ATRYUM_TOKEN_COMMAND or ATRYUM_URL invalidates the cache instead
+// of sending a token minted for a different identity or target.
+const TOKEN_CACHE_KEY = TOKEN_COMMAND
+  ? createHash("sha256").update(`${TOKEN_COMMAND}\n${API}`).digest("hex")
+  : "";
+let cachedToken = ACCESS_TOKEN;
```
hunner: Nit: `cachedToken = ACCESS_TOKEN` with expiry `0` when `TOKEN_COMMAND` is set is dead state — the token can never be returned from the cache check. Initializing to `""` when `TOKEN_COMMAND` is set would be less misleading.

🤖 _Automated reply_

examples/amp-plugin/atryum.ts
```diff
@@ -65,14 +73,172 @@ const CLIENT_VERSION =
...
+const STATE_DIR =
+  process.env.ATRYUM_STATE_DIR ||
+  join(homedir(), ".atryum", "amp-plugin-state");
+const TOKEN_CACHE_FILE = TOKEN_COMMAND
+  ? join(STATE_DIR, "token-cache.json")
+  : "";
+// Ties the cached token to the command and server that produced it, so
+// switching ATRYUM_TOKEN_COMMAND or ATRYUM_URL invalidates the cache instead
+// of sending a token minted for a different identity or target.
+const TOKEN_CACHE_KEY = TOKEN_COMMAND
```
hunner, line 95: Nit: the cache key covers the command *text* but not env vars it references — changing `$CLIENT_ID` without changing the command string reuses the old token until expiry. Probably fine, but a one-line mention in the docs' cache-key sentence would save someone a confused debugging session.

🤖 _Automated reply_

hunner (approved):
Solid, well-tested feature with unusually careful cache handling — the fresh-temp-file + atomic rename dance so `0600` actually applies, the sha256(command + URL) cache keying, the single-flight refresh promise, and the bounded one-shot 401 retry are all right. Enabling Pi in the integration harness so the full mint → cache → re-mint → 401-retry path is exercised end-to-end is great. Docs are accurate to the code (I verified the 55-minute default, skew, cache keying, mode, and retry semantics all match).

Approving with five issues and three nits inline — none blocking, but 1, 2, and 5 are small concrete fixes worth making before merge.

🤖 _Automated reply_

mdeyell-valid-mind:
@hunner Thanks, I updated this

hunner (approved):
🥟

PR #132, opened 2026-07-13
Refactor rule creation UX: pre-filled modal for always approve/deny

even-steven:
## What and why?

A collection of fixes and UX improvements to the Agent Authority invocations and rules experience.

---

### 1. Show AI evaluation reasons in audit display

Instead of a generic "Deferring to human approval as per charter" label, the audit trail now shows the actual LLM-generated reason from the AI evaluation (e.g. the specific charter clause or rule that caused deferral or approval).

### 2. Fix invocation audit display for no-match and deleted rules

- Invocations with no matched rule now show **\*No rule matched\*** (previously blank or inconsistent)
- Invocations matched to a rule that has since been deleted show **\*Deleted Rule\*** instead of "No rule matched"
- "Invocation approved/denied by AI" vs "by rule" vs "by server default" now correctly differentiated
- AI Evaluation badge hidden for server-default fallback events

### 3. Fix SSE updates not refreshing the invocation list

After pagination was added, the SSE stream was writing to a cache key without an `offset`, so updates were invisible to the paginated view. Fixed by switching to `invalidateQueries` so the active page always refetches.

### 4. Refactor rule creation UX: pre-filled modal for always approve/deny

Replaces the implicit "always approve / always deny" behaviour — which silently created an overly-broad rule matching every server, tool, and agent — with a pre-filled **Create Rule** modal that lets the user review and narrow the scope before saving.

- `CreateRuleModal` extracted from `Rules.tsx` → `ui/src/components/CreateRuleModal.tsx`
- "Always approve" / "Always deny" open the modal pre-filled with `auto_approve` / `auto_deny` + the invocation's server, tool, and agent
- "Create Rule From This" replaced with the same `CreateRuleModal`
- Rule description auto-generated from action + server + tool (e.g. "Auto-approve filesystem/read_file")
- Rules cache invalidated on every SSE event so newly created rules resolve immediately without a hard reload

## How to test

1. Trigger an invocation evaluated by AI — audit trail should show the actual LLM reason, not a generic string
2. Trigger an invocation with no matching rule — should show "*No rule matched*" with no badge
3. Delete a rule, then view an invocation that previously matched it — should show "*Deleted Rule*"
4. On the Invocations page, click "Always approve" → modal opens pre-filled with server/tool/agent and a generated description; adjust and save → rule appears in Rules list immediately
5. Repeat with "Always deny"
6. Create a rule while on the Invocations page, trigger a matching invocation — audit display should show the rule name without a hard reload

## What needs special review?

- The inline "Customize rule scope" UI and old inline Create Rule modal are fully removed
- Rules cache is now invalidated on every SSE tick — may cause extra network requests at high invocation volume

Companion PR in `validmind/frontend`: https://github.com/validmind/frontend/pull/2724

## Release notes

Invocation audit trails now show AI evaluation reasons and correctly distinguish deleted rules from unmatched ones. The "Always approve / Always deny" actions now open a pre-filled rule editor instead of silently creating broad rules.

## Checklist
- [x] What and why
- [ ] Screenshots or videos
- [x] How to test
- [x] Labels applied
- [x] Tested locally

### DIFF
--- a/ui/src/pages/Invocations.tsx
+++ b/ui/src/pages/Invocations.tsx
@@ -22,13 +22,6 @@ import {
   MenuButton,
   MenuItem,
   MenuList,
-  Modal,
-  ModalBody,
-  ModalCloseButton,
-  ModalContent,
-  ModalFooter,
-  ModalHeader,
-  ModalOverlay,
   Spinner,
   Stack,
   Table,
@@ -45,6 +38,7 @@ import {
   Alert,
   AlertDescription,
   AlertIcon,
+  useDisclosure,
   useToast,
 } from "@chakra-ui/react";
 import { Select } from "chakra-react-select";
@@ -56,6 +50,7 @@ import {
 } from "@heroicons/react/24/outline";
 
 import { ContentPageTitle } from "../components/Layout";
+import { CreateRuleModal } from "../components/CreateRuleModal";
 import DiffViewer from "../components/DiffViewer";
 import { extractDiff } from "../components/diffUtils";
 import AgentIcon, { detectAgentKind } from "../components/AgentIcon";
@@ -68,7 +63,7 @@ import {
   useSummarizeInvocation,
 } from "../hooks/useInvocations";
 import { useSettings } from "../hooks/useSettings";
-import { useCreateRule, useRules } from "../hooks/useRules";
+import { useRules } from "../hooks/useRules";
 import { useAgents } from "../hooks/useAgents";
 import { useInvocationStream } from "../hooks/useInvocationStream";
 import {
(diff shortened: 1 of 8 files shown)


ui/src/pages/Invocations.tsx
mdeyell-valid-mind: i think this is dead code

mdeyell-valid-mind:
When someone makes a new rule in this modal, it should reevaluate the current invocation

ui/src/pages/Invocations.tsx
even-steven: Good catch — this is dead code left over from the old "apply rule to all pending invocations" flow. Removed in 7bcc9c3 along with the now-unused `Invocation` type import.

mdeyell-valid-mind (approved):
i didn't run this but the code looks good

mdeyell-valid-mind (approved):
approve again

Issue #90, opened 2026-06-16
Agent plan pre-approval


hunner:
Support higher-level plan submission where an agent proposes a sequence of intended actions with rationale, for example: run commands a, b, c to achieve x. Atryum returns approve, revise, or deny before any tool executes. Enables batch intent review instead of per-call gating only.

mdeyell-valid-mind:
The pre approval plan will be optional not required

PR #133, opened 2026-07-13
Agent plan preapprove

mdeyell-valid-mind:
We would like to give agents the option to submit plans for preapproval. Similar to normal invocations rules can be applied to plans.
## Summary

  Adds agent plan preapproval, allowing agents to submit plans for approval before executing their individual tool calls.

  Plan approval uses the existing invocation rules. Each planned action is evaluated against relevant rules in order, while a shared LLM judge reviews the entire plan for charter
  compliance at most once.

  ## What changed

  - Added plan submission, approval, rejection, cancellation, expiration, revision, and completion workflows.
  - Added plan persistence, API endpoints, MCP tools, SSE events, and UI pages.
  - Added plan support to the Amp, Claude Code, Codex, Cursor, and Pi integrations.
  - Added plan notifications and links between plans and their invocations.

  ## Plan approval behavior

  For each action in a submitted plan, relevant invocation rules run from top to bottom:

  - auto_deny rejects the entire plan immediately.
  - auto_approve approves that action.
  - ai_evaluation requires whole-plan charter review.
  - Human-approval rules are treated as preapproved once the plan is approved.
  - No matching rule falls back to human plan approval.

  A mandatory LLM judge reviews the entire plan for charter compliance:

  - It runs at most once per plan, even when multiple actions trigger AI-evaluation rules.
  - It may be skipped when an invocation rule deterministically denies the plan.
  - A noncompliant plan is rejected.
  - A missing charter is rejected.
  - Evaluator or configuration failures fall back to human approval.
  - An LLM revise result cannot override an auto_deny.
  - Empty local evaluator configuration now correctly uses the default enabled configuration.

  ## Plan execution behavior

  When an approved plan runs:

  - Invocations are matched to the newest applicable active plan.
  - Current auto_deny rules still block execution.
  - Rules already satisfied during plan approval are treated as preapproved.
  - An adherence judge verifies that the concrete invocation arguments follow an approved plan action and the charter.
  - Steps cannot execute backwards after a later step has run, though skipping forward is allowed.
  - Duplicate candidate actions bind to the earliest eligible step.
  - Status polling does not advance plan progress.
  - Completing the final successful action completes the plan.

  The implementation supports both authenticated and no-auth deployments, including managed-agent identities and aliases.

  ## UI

  - Added the Plans page and plan detail views.
  - Added human approval and rejection controls.
  - Added live SSE updates and approval notifications.
  - Added navigation between plans and related invocations.

  ## Testing

  Added coverage for:

  - Invocation-rule and judge decision combinations.
  - Multiple AI rules sharing exactly one judge call.
  - Auto-deny precedence over judge revisions.
  - Deterministic auto-deny skipping the judge.
  - Auto-approve plans without a charter being rejected.
  - Evaluator failure and default LLM configuration handling.
  - Authenticated and no-auth charter resolution.
  - Plan ordering, adherence, ownership, expiration, and completion.
  - Plan API, UI, SSE, and harness integrations.

https://github.com/validmind/atryum/issues/90
Screenshot 2026-07-13 at 11 47 50
Screenshot 2026-07-13 at 11 47 34

Screenshot 2026-07-13 at 11 48 59

Note that because claude managed agents doesn't have a harness it won't get the hook telling it about this feature.


internal/invocation/plan_service.go
```diff
@@ -0,0 +1,776 @@
...
+		for _, a := range plan.Actions {
+			if a.Tool == tool && (a.Server == "" || a.Server == server) {
+				return approvedPlanMatch{Plan: plan, Action: a}, true
+			}
+		}
+	}
+	return approvedPlanMatch{}, false
+}
+
+func (s *Service) approvedPlanPass(ctx context.Context, match approvedPlanMatch, agentRec AgentRecord, server, tool string, input map[string]any, extraContext string) (string, *float64, bool) {
```
hunner: **Fix requested — run the adherence judge on every pass, not just AI-approved plans.**

The adherence judge only runs when the plan's matched rule was `ai_evaluation`. Human-approved plans (the default no-rule path) and `auto_approve`-rule plans grant a deterministic pass on tool name alone — so no one, human or judge, ever sees the concrete arguments of any call executed under the pass. `input_summary`/`description` are explicitly non-binding (per the `PlanAction` doc comment), which means an action like `tool=Bash` authorizes arbitrary shell for the TTL once a human waves the plan through.

This also creates an incentive gradient: vague, benign-sounding plans are easier to get approved *and* buy a broader pass, so the agent (or a prompt-injected agent) is rewarded for under-describing. Binding adherence checks remove that gradient — the concrete call gets compared against the plan's own summaries regardless of who approved it.

Suggestion: run `EvaluatePlanAdherence` for every pass whenever an LLM config is available (the status-poll fast pass still applies), not only when `plan.MatchedRuleID` points at an `ai_evaluation` rule.

🤖 _Automated reply_

internal/invocation/local_evaluator.go
```diff
@@ -249,6 +249,277 @@ func (e *LocalEvaluatorClient) parseVerdict(raw string) (verdict string, confide
...
+		// valid verdicts
+	default:
+		slog.Warn("local plan evaluator: unrecognised verdict; falling back to denied", "verdict", out.Verdict)
+		out.Verdict = "denied"
+	}
+	c := out.Confidence
+	return PlanEvaluateResponse{Verdict: out.Verdict, Reason: out.Reason, Feedback: out.Feedback, Confidence: &c}
+}
+
+const planAdherenceSystemPrompt = `You are an AI safety judge. Evaluate whether an actual tool call follows an already-approved plan and the specific planned action it matched.
```
hunner: **Fix requested — give the adherence judge a charter-violation path.**

The verdict set (`follows_plan` / `outside_plan` / `human_approval`) frames the question purely as plan-conformance. The charter is in the prompt, but there is no verdict for "this call plausibly follows the plan, yet its concrete arguments violate the charter" — exactly the case the plan judge could not catch, since it never saw real arguments. A call matching a vaguely-worded action can legitimately return `follows_plan` while doing something the charter prohibits.

Suggestion: add an explicit instruction along the lines of "even if the tool call follows the approved plan, return outside_plan or human_approval if its concrete arguments violate the charter", plus a test exercising that case.

🤖 _Automated reply_

internal/invocation/plan_service.go
```diff
@@ -0,0 +1,776 @@
...
+		slog.Warn("plan pass lookup failed; falling back to normal gating", "agent_id", agentID, "error", err)
+		return approvedPlanMatch{}, false
+	}
+	for _, plan := range plans {
+		plan = s.expireIfStale(ctx, plan)
+		if plan.Status != PlanStatusApproved {
+			continue
+		}
+		for _, a := range plan.Actions {
+			if a.Tool == tool && (a.Server == "" || a.Server == server) {
```
hunner: **Fix requested — default an omitted `action.server` to the submitting source.**

`a.Server == ""` matches a same-named tool on *any* connected upstream, while `matchPlanRules` only matches rule `ServerPatterns` against `plan.Source`. So a plan judged/approved in the context of one server grants a pass to any tool sharing the name on any other server (tool-name collisions across MCP servers are common: `create_issue`, `search`, ...).

Suggestion: stamp an omitted `action.server` with the submitting source at `SubmitPlan` time — mirroring how `handleMCPPlanSubmit` already defaults `Source` to the MCP server. Every stored action becomes concrete, the wildcard stops being representable, and cross-server scope only exists when explicitly declared (and therefore visible to the plan judge and the human reviewer). Requiring the field instead would add friction and a rejection case for callers that can't easily know Atryum's server naming.

🤖 _Automated reply_

internal/store/plans.go
```diff
@@ -0,0 +1,321 @@
...
+	if err != nil {
+		return err
+	}
+	_, err = r.db.ExecContext(ctx, query, args...)
+	return err
+}
+
+// Update persists the mutable review fields of a plan: status, approval,
+// feedback, matched rule, expiry, and decision time.
+func (r *PlansRepo) Update(ctx context.Context, p invocation.Plan) error {
```
hunner, line 59: **Bug — `TTLSeconds` is not persisted on update.**

`ApprovePlan` with a reviewer TTL override sets `plan.TTLSeconds` and `finalizePlanDecision` computes `ExpiresAt` from it, but this `Update` never writes `ttl_seconds`. The persisted `expires_at` is correct, so expiry behaves — but the `ttl_seconds` column keeps the agent's originally-requested value, and anything displaying TTL from the DB (UI, API responses after a reload) will lie.

🤖 _Automated reply_

hunner (commented):
Solid, carefully-engineered implementation — the fail-closed plumbing (parse failures → denied/human, LLM errors → human review or normal gating, missing charter → deny), the status-poll fast pass with its adversarial test matrix, and the overall test discipline are genuinely good. The concerns below are at the policy-design layer, not the implementation.

The through-line of the inline comments: the plan path is not judge-equivalent to unplanned execution. The plan judge necessarily sees agent-authored summaries rather than concrete arguments (inherent to preapproval), so the adherence judge is the mechanism that holds execution to the approved intent — but it currently only runs for AI-approved plans, its verdict set can't catch charter violations that follow the plan, and the judge/reviewer are deciding under a one-shot mental model of a pass that is actually unlimited-use until expiry.

Summary of the comments:

**Fixes requested**
1. Run the adherence judge on human- and auto-approved plans too (`approvedPlanPass`)
2. Give the adherence judge a charter-violation path (`planAdherenceSystemPrompt`)
3. Default an omitted `action.server` to the submitting source instead of matching any upstream
4. Persist `TTLSeconds` in `PlansRepo.Update`
5. Test coverage for the `Invoke` plan-pass branch (and dedupe with `Submit`)

**Discussion starters**
6. Should approved-plan passes be consumable/counted rather than unlimited-use within TTL? (`planJudgeSystemPrompt`)
7. Should invocation-scoped `auto_deny` rules stay authoritative over plan passes?

🤖 _Automated reply_

internal/invocation/service.go
```diff
@@ -1276,6 +1321,33 @@ func (s *Service) Submit(ctx context.Context, req ExternalSubmitRequest) (Invoca
 	if sessionID != "" {
 		_ = s.sessions.TouchSession(ctx, sessionID, now.Add(externalSessionTTL)) // best-effort last-seen/expiry bump
 	}
+
+	// An approved plan grants a scoped pass: a declared action matching this
+	// call auto-approves it before rule matching, recording the plan as the
+	// approval reason.
+	if planMatch, ok := s.matchApprovedPlan(ctx, agentID, source, req.Tool); ok {
```
hunner: **Tests — the `Invoke` plan-pass branch is untested.**

All pass-grant/adherence tests exercise `Submit`; the equivalent branch in `Invoke` (~line 350) is a ~40-line near-duplicate with subtly different event emission (`invocation.executing` + `finishExecution` vs `StatusApproved` + `invocation.approved`) and passes `""` as the adherence judge's context where this path passes `sessionContext`. Extracting a shared helper would remove the duplication and make both paths testable with the same table.

🤖 _Automated reply_

internal/invocation/local_evaluator.go
```diff
@@ -249,6 +249,277 @@ func (e *LocalEvaluatorClient) parseVerdict(raw string) (verdict string, confide
 	return out.Verdict, out.Confidence, out.Reason
 }
 
+const planJudgeSystemPrompt = `You are an AI safety judge. An agent has proposed a plan — a goal, a rationale, and a sequence of intended tool calls — and asks for approval before executing any of them. Evaluate the plan as a whole against the agent's governing charter.
```
hunner, line 256: **Discussion — the judge (and reviewer) decide under one-shot semantics, but the pass is unlimited-use.**

The prompt describes "a sequence of intended tool calls", but approval actually grants an unlimited-use, order-independent pass per tool name until `expires_at` — actions are never consumed by `matchApprovedPlan`. A judge approving "delete this one temp file" doesn't know it's approving unbounded repeats for up to 24h. The human approval screen has the same gap.

Is this a gap we should close rather than just disclose? Options, roughly in order of strictness:

- **(a) One-shot consumption** — each action is consumed when matched. Strictest, but breaks legitimate workflows that repeat a tool (several `Edit` calls, retry after a transient failure).
- **(b) Per-action use counts** — the plan declares an expected count (defaulting to something small), and the pass stops matching once exceeded. Keeps the repeat-friendly ergonomics while bounding blast radius; probably the best cost/benefit.
- **(c) Keep unlimited, disclose** — state the real grant semantics in this prompt and in the Plans approval UI so both judges and humans decide with the right model.

(c) seems like the floor either way; worth deciding (a)/(b) before these semantics ossify into the API.

🤖 _Automated reply_

internal/invocation/service.go
```diff
@@ -340,6 +344,47 @@ func (s *Service) Invoke(ctx context.Context, req CreateInvocationRequest) (Invo
 		return InvocationResponse{}, err
 	}
 
+	// An approved plan grants a scoped pass: a declared action matching this
+	// call executes immediately without rule/policy gating, recording the
+	// plan as the approval reason.
+	if planMatch, ok := s.matchApprovedPlan(ctx, agentID, upstream.Name, req.Tool); ok {
```
hunner: **Discussion — should `auto_deny` rules stay authoritative over plan passes?**

The plan pass runs *before* rule matching (here and in `Submit`), so an approved plan overrides not just `human_approval` friction but explicit invocation-scoped `auto_deny` rules. That's consistent with the PR description ("bypassing existing rules"), but an admin who wrote a deny rule almost certainly expects it to be absolute — deny rules read as policy, not as friction to be waived.

Two ways to resolve it:
1. Check matching `auto_deny` invocation rules before granting the pass, so deny always wins; or
2. Keep the full bypass, but document it prominently (rules UI + README), since it changes what a deny rule means once plan-scoped rules exist.

Leaving this as a decision point for the team — flagging it so it's chosen deliberately.

🤖 _Automated reply_

internal/invocation/local_evaluator.go
hunner: Yeah so it sounds like a plan is "do X Y and Z" but it actually allows doing X Y and Z any number of times and in any order until the TTL. With the secondary plan-adherence judge still evaluating, everything must still come under the plan (and under the charter after the above bugfix), but this may be considered a gap.

I know agents sometimes take multiple tries to get a command right, so maybe we can have order dependency (skipping steps should be allowed) and limit for the number of times that each step can be re-tried (five times or whatever)?

internal/invocation/service.go
hunner: It seems like `auto_deny` rules, even ones at the very top of the rule list getting skipped might be an issue. Those rules may not be mentioned in charters, since under normal circumstances they would always apply and not need to be explicitly mentioned, but in that case a judge could approve a plan and then end up with tool calls that would otherwise have been denied by static rules.

And judges shouldn't need the rules, since the only things that they would have to judge on normally will be the things that are allowed by the rules list.

hunner: Funnily enough, the agent CAN see the rules so has better visibility into the governance than the judge doing the governing. With this gap the agent could compose plans that route around rules that the judge can't see.

examples/shared-agent-hook/atryum-hook.mjs
```diff
@@ -293,6 +294,204 @@ function describe(input) {
...
+  if (!res.ok) return undefined;
+  return res.json();
+}
+
+async function planHint(tool) {
+  planSupport ||= loadAgentRules(tool).catch(() => undefined);
+  const rules = await planSupport;
+  if (!rules?.plan_submission?.enabled) return "";
+  const endpoint = rules.plan_submission.endpoint || "/api/v1/external/plans";
+  return ` Atryum supports preapproval plans for complex, risky, or multi-step work. For tasks with two or more tool calls, file changes, shell commands, external systems, or ordered actions, submit a batch plan to ${endpoint} before running tools, then wait for approval before executing the planned steps.`;
```
hunner: this can account for database-transaction like command-joining safety so that it doesn't end up in broken state part way through due to a denied tool call

internal/invocation/service.go
hunner: We can run every command in a plan through all deny and human approval rules.

internal/invocation/local_evaluator.go
hunner: yeah sounds like in-order and limited number of retries should be fine

internal/invocation/local_evaluator.go
hunner: this is actually intentional, since the judge should evaluate the original plan against the charter, and thus the adherence judges shouldn't need to deny based on charter. And perhaps the adherence judge shouldn't even have the charter in context?

mdeyell-valid-mind:
I talked to @hunner about this in a meeting and am putting it back in draft while i revise it

even-steven (commented):
I know @nibalizer had brought this up with one of his demos that Invocations were not displaying in real-time, so I'm assuming the same is true for Plans

Invocations → SSE (server-push). useApprovalNotifications opens a streaming connection to /api/v1/admin/invocations/stream?status=pending_approval with Accept: text/event-stream, reads it via response.body.getReader(), and reacts to named invocations events as the server pushes them. Pending invocations therefore surface in near real-time.

Plans → polling (client-pull). There's no plan stream endpoint. Two independent 5-second setTimeout loops fetch /api/v1/admin/plans?status=pending_approval:

one in useApprovalNotifications (for the desktop notification), and
one in usePlans via react-query's refetchInterval (for the Plans page).
The hook author even flags it explicitly in usePlans.ts: // Plans have no SSE stream — poll every 5s so pending submissions show up.

mdeyell-valid-mind:
> I know @nibalizer had brought this up with one of his demos that Invocations were not displaying in real-time, so I'm assuming the same is true for Plans
> 
> Invocations → SSE (server-push). useApprovalNotifications opens a streaming connection to /api/v1/admin/invocations/stream?status=pending_approval with Accept: text/event-stream, reads it via response.body.getReader(), and reacts to named invocations events as the server pushes them. Pending invocations therefore surface in near real-time.
> 
> Plans → polling (client-pull). There's no plan stream endpoint. Two independent 5-second setTimeout loops fetch /api/v1/admin/plans?status=pending_approval:
> 
> one in useApprovalNotifications (for the desktop notification), and one in usePlans via react-query's refetchInterval (for the Plans page). The hook author even flags it explicitly in usePlans.ts: // Plans have no SSE stream — poll every 5s so pending submissions show up.

Excellent feedback, thanks

even-steven (approved):
👍🏽

hunner (approved):
🗓️

Nice work!

The only interesting thing I saw is that if a plan has actions that are 100% covered by `auto_approve` rules, `plan_service` still runs the `evaluateMandatoryPlanCharter` against the plan. This means an agent without a charter can run auto-approved commands, but would get a plan denied for the lack of a charter. Just a weird gap that doesn't really matter in real scenarios I don't think.

PR #136, opened 2026-07-14
Central version injected at release time, exposed via `atryum version`

hunner:
## What this does

Gives atryum a single source of truth for its version instead of `0.1.0` hardcoded in six places (which had drifted — every binary since v0.0.x has claimed `0.1.0`).

- `internal/version.Version` defaults to `"dev"`; `release-build` injects the release tag via `-ldflags "-X atryum/internal/version.Version=<tag>"`, so releases self-version and nothing needs bumping by hand.
- MCP handshake metadata reports it: `serverInfo.version` (what connecting MCP clients see) and all four `clientInfo.version` sites (what upstream MCP servers see).
- New `atryum version` command (also `--version`/`-v`), which prints the version alongside the VCS revision go stamps into the binary: `dev (d09346abf30e, modified)` for a local build, `v0.2.0 (<tag commit>)` for a release.
- `release-build` asserts the freshly built host-platform binary self-reports the tag — `-ldflags -X` silently no-ops if the version symbol is ever renamed, and this turns that into a loud release failure instead of binaries quietly shipping as `dev`.
- `ui/package.json` and `package-lock.json` pinned to `0.0.0` — nothing consumes the npm version, and an out-of-sync lock file would let `npm install` dirty the pristine release clone and flip the binaries' `vcs.modified` stamp to `true`.

## Verification

`go build ./...` and tests for the touched packages pass. Exercised both build flavors: a dev build prints `dev (<revision>, modified)`; a full `just release-build` from a pristine clone with a throwaway tag produced binaries reporting `<tag> (<revision>)` with `vcs.modified=false`, and the self-report assertion passed. Negative case verified too: building a tag whose source predates `internal/version` fails the release as intended.

### DIFF
--- a/cmd/atryum/main.go
+++ b/cmd/atryum/main.go
@@ -48,6 +48,8 @@ func main() {
 		err = runHooks(os.Args[2:])
 	case "licenses":
 		err = runLicenses()
+	case "version", "--version", "-v":
+		fmt.Println(versionString())
 	default:
 		err = fmt.Errorf("unknown command %q\n%s", os.Args[1], globalUsage())
 	}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Release builds run from a pristine clone of the release tag, with the
   tag's commit stamped into the binaries as VCS provenance, and GitHub
   release notes are drawn from this changelog.
+- `atryum version` command (also `--version`). Release builds report the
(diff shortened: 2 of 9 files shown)


cmd/atryum/main.go
```diff
@@ -48,6 +48,8 @@ func main() {
 		err = runHooks(os.Args[2:])
 	case "licenses":
 		err = runLicenses()
+	case "version", "--version", "-v":
+		fmt.Println(versionString())
```
mdeyell-valid-mind, line 52: 👍

mdeyell-valid-mind (approved):

PR #139, opened 2026-07-16
initial arch docs

kam-validmind:
A first iteration of architecture docs


mdeyell-valid-mind (approved):
would it be worth adding something about how rule matching works?

PR #143, opened 2026-07-21
Fix: ClearSessions stops each watcher

kam-validmind:
# Pull Request Description

## What and why?

**Bug fix:** `Service.ClearSessions` (`internal/managedagents/service.go`) could leave background
"watcher" goroutines running for sessions it had already deleted from the database, if a delete
partway through the batch failed.

**Background for context.** Every registered Claude Managed Agents session has one background
goroutine ("watcher") polling Anthropic and acting on that session's tool calls. Atryum has no
other gate on those tool calls — the watcher *is* the approval mechanism (Anthropic pauses the
agent and waits for the watcher's allow/deny). Stopping a watcher means calling its
`context.CancelFunc` from the in-memory `watchers map[string]context.CancelFunc`; if nothing ever
calls `cancel()`, the goroutine just keeps running until the process exits.

`ClearSessions` is supposed to do two things, together, for every session: delete its DB row and
cancel its watcher. Before this fix it did them in two entirely separate loops over the same list —
delete everything first, *then* go back and cancel everything:

```go
for _, session := range sessions {                 // Loop 1: delete every session
    if err := s.sessions.Delete(ctx, session.SessionID); err != nil {
        return 0, err                               // bails out of the WHOLE function
    }
}
s.mu.Lock()
for _, session := range sessions {                 // Loop 2: stop every watcher
    if cancel, ok := s.watchers[session.SessionID]; ok {
        cancel()
        delete(s.watchers, session.SessionID)
    }
}
s.mu.Unlock()
return len(sessions), nil
```

If `Delete` failed for any single session, the function returned immediately — and loop 2 (stopping
watchers) never ran at all, not even for the sessions loop 1 had already deleted successfully.

**Before this fix** (5 sessions, the 3rd delete hits a transient DB error):

```mermaid
sequenceDiagram
    autonumber
    participant Admin
    participant Clear as Service.ClearSessions
    participant DB as SessionStore
    participant W as watchers map (in memory)

    Admin->>Clear: "Clear all sessions" (5 sessions)
    Clear->>DB: List()
    DB-->>Clear: [sess_1 .. sess_5]
    Clear->>DB: Delete(sess_1)
    DB-->>Clear: OK
    Clear->>DB: Delete(sess_2)
    DB-->>Clear: OK
    Clear->>DB: Delete(sess_3)
    DB-->>Clear: err (transient DB error)
    rect rgb(120, 30, 30)
    Note over Clear: BUG: returns the error right here.<br/>Loop 2 (stop watchers) never runs — not even once.
    end
    Clear-->>Admin: error, "0 cleared"
    Note over W: watchers["sess_1"] and watchers["sess_2"]<br/>are STILL in the map, STILL running —<br/>polling Anthropic and acting on tool calls<br/>for sessions that no longer exist in the DB
```

Why this matters more than "some harmless background polling": the zombie watcher's tool calls
still go through the exact same real approval pipeline as any legitimate call (rule matching,
policy, human review) — nothing about them looks wrong from inside Atryum, and nothing on
Anthropic's side can tell a zombie watcher apart from the intended one for that session (both hold
the same API key). An admin who clears sessions to fully unlink/decommission an agent can be left
with a fully-authorized decision-maker for that agent still running, invisible in the session list,
until the process restarts. It is **not** a security-boundary bypass — the zombie still runs every
rule/policy/human-review step before it ever sends an allow/deny — but it is a real operational gap
in Atryum's own bookkeeping of what it's currently governing. Full write-up, including why the gate
mechanism makes this specifically dangerous and a table of what an admin can/can't observe or do
about it: **`docs/bug-clear-sessions-zombie-watchers.md`**.

**The fix:** merge the two loops into one — cancel each session's watcher **immediately after**
its own delete succeeds, so a failure partway through can only ever leave the *not-yet-reached*
sessions untouched, never a session that was already deleted with its watcher still running. This
also makes `ClearSessions` consistent with its single-session sibling, `DeleteSession`, which
already deletes-then-cancels correctly per call.

```go
cleared := 0
for _, session := range sessions {
    if err := s.sessions.Delete(ctx, session.SessionID); err != nil {
        return cleared, err          // partial progress so far, honestly reported
    }
    s.mu.Lock()
    if cancel, ok := s.watchers[session.SessionID]; ok {
        cancel()
        delete(s.watchers, session.SessionID)
    }
    s.mu.Unlock()
    cleared++
}
return cleared, nil
```

**After this fix** (same failure):

```mermaid
sequenceDiagram
    autonumber
    participant Admin
    participant Clear as Service.ClearSessions
    participant DB as SessionStore
    participant W as watchers map (in memory)

    Admin->>Clear: "Clear all sessions" (5 sessions)
    Clear->>DB: List()
    DB-->>Clear: [sess_1 .. sess_5]
    Clear->>DB: Delete(sess_1)
    DB-->>Clear: OK
    Clear->>W: cancel(sess_1) + remove from map
    Clear->>DB: Delete(sess_2)
    DB-->>Clear: OK
    Clear->>W: cancel(sess_2) + remove from map
    Clear->>DB: Delete(sess_3)
    DB-->>Clear: err (transient DB error)
    rect rgb(30, 90, 30)
    Note over Clear: FIXED: returns immediately, same as before —<br/>but sess_1/sess_2's watchers are already stopped
    end
    Clear-->>Admin: error, "2 cleared" (was always "0" before)
    Note over W: No zombies: sess_1 and sess_2 are gone from<br/>both the DB and the watchers map.<br/>sess_3..sess_5 untouched by either step (as expected).
```

Two changes worth calling out explicitly:

- **Delete and cancel now happen inside the same loop iteration**, back to back — this is the
  actual fix. It's the exact per-session ordering `DeleteSession` already used correctly;
  `ClearSessions` just hadn't applied it across a loop.
- **The error-path return value changed from a hardcoded `0` to `cleared`.** Previously, even if 2
  of 5 sessions were genuinely removed before the failure, the caller was told `0`. The service
  layer now reports how much progress actually happened.

The happy path (no delete fails) is behaviorally identical: same sessions deleted, same watchers
stopped, same `len(sessions)` returned.

**Files changed:**
- `internal/managedagents/service.go` — merged loop, per-session lock scope, honest partial count
- `internal/managedagents/service_test.go` — new regression test (new file)

## How to test

Automated — new regression test, verified to fail before the fix and pass after (confirmed by
reverting the fix locally and re-running):

```bash
go test ./internal/managedagents/ -run TestClearSessionsStopsWatchersForSessionsDeletedBeforeAFailure -v
```

`TestClearSessionsStopsWatchersForSessionsDeletedBeforeAFailure` builds a fake `SessionStore` whose
`Delete` fails only for one session ID (`sess_3`) in the middle of a 5-session list, pre-populates
`svc.watchers` with tracking `cancel` funcs, calls `ClearSessions`, and asserts the watchers for the
two sessions that *did* delete successfully (`sess_1`, `sess_2`) were actually stopped. Output
before the fix:

```
=== RUN   TestClearSessionsStopsWatchersForSessionsDeletedBeforeAFailure
    service_test.go:81: watcher for sess_1 was not stopped despite its DB row being deleted (zombie watcher)
    service_test.go:81: watcher for sess_2 was not stopped despite its DB row being deleted (zombie watcher)
--- FAIL: TestClearSessionsStopsWatchersForSessionsDeletedBeforeAFailure (0.00s)
FAIL
```

Now fixed and passing. Also ran the full suite, `go vet`, and `gofmt` to confirm nothing else
regressed:

```bash
go test ./...
go vet ./...
gofmt -l internal/managedagents/service.go internal/managedagents/service_test.go
```

All pass locally / produce no output (`gofmt`) as of this PR.

## What needs special review?

- **Lock granularity change.** The old code took `s.mu` once for the entire "stop every watcher"
  loop; the fix now locks/unlocks `s.mu` once per iteration, interleaved with the (unlocked)
  `s.sessions.Delete` DB call for the *next* session. This is a narrower critical section (matches
  `DeleteSession`'s existing pattern) and I believe it's safe, but please double check there's no
  path where a concurrent `RegisterSession`/`DeleteSession` call interleaving between iterations
  could observe or produce a surprising intermediate state in `s.watchers`.
- **The `cleared` count doesn't reach the HTTP caller on error today.** The only caller,
  `internal/api/handlers.go:3510` (`DELETE` on the sessions endpoint), does
  `deleted, err := h.managedAgents.ClearSessions(...)` and on any `err != nil` returns
  `500 {err.Error()}` — it never reads `deleted` in that branch, only on the full-success path via
  `ManagedAgentSessionClearResponse{Deleted: deleted}`. So this fix makes the service layer report
  an honest partial count, but an admin hitting a partial failure today still only sees a bare error
  string, not "2 of 5 cleared." Flagging as a known gap, intentionally **out of scope** here (would
  be a response-shape change on the admin API, not just an internal fix) — worth a follow-up ticket
  if partial-progress visibility matters to admins.
- **Related, narrower bug in `DeleteSession` — also out of scope.** Deleting a *single* session
  applies the same delete-then-cancel order, so if a session's DB row is already missing (e.g. it
  was already zombied by this exact bug before the fix, or by a retry), that single delete-and-cancel
  fails with "not found" and — same shape as this bug — bails out before reaching the cancel step.
  Worth its own fix, but `ClearSessions` is the only path that *creates* the zombie state in the
  first place, so it's the one fixed here. See "Related gap, not fixed here" in
  `docs/bug-clear-sessions-zombie-watchers.md`.
- **The valid/happy path is unchanged.** When every delete succeeds, `cleared == len(sessions)`,
  identical to the old `return len(sessions), nil`. A reviewer should confirm the diff really is
  just a reordering plus an honest count, not a behavior change on the success path.

- No breaking changes to the public API contract: the success-path response shape
  (`{"deleted": N}`) is byte-identical to before. The only observable change is that the
  *internal* `ClearSessions(ctx)` return value on error is now the true partial count instead of a
  hardcoded `0` — not currently surfaced by the one HTTP caller (see review note above), so no
  client-visible behavior changes from this PR alone.
- No special deployment considerations — plain application code, no data migration required.
  Rolling deploy is fine.

## Data impact

Justification: this only reorders existing DB-delete and in-memory-cancel operations within a
single function and corrects an internal return value on an error path. It does not modify any
customer data, configuration format, or the shape of any API response.

## Release notes

Label: `bug`

Fixed an issue where clearing all managed-agents sessions could leave background watcher processes
running for sessions that were already deleted, if a database error interrupted the clear
operation partway through. Each session's background watcher is now stopped immediately after its
own deletion succeeds, so a partial failure can no longer leave orphaned watchers running.

## Checklist
- [x] What and why
- [x] How to test
- [x] Data impact classified
- [x] Tested locally
- [x] Documentation updated (if required) — see `docs/bug-clear-sessions-zombie-watchers.md`


internal/managedagents/service_test.go
```diff
@@ -0,0 +1,84 @@
...
+	for _, sess := range store.sessions {
+		id := sess.SessionID
+		svc.watchers[id] = func() {
+			mu.Lock()
+			stopped[id] = true
+			mu.Unlock()
+		}
+	}
+
+	if _, err := svc.ClearSessions(context.Background()); err == nil {
```
mdeyell-valid-mind: Could we capture and assert the returned count is 2 here?

mdeyell-valid-mind (approved):

internal/managedagents/service_test.go
kam-validmind: good idea, now addressed

mdeyell-valid-mind (approved):

Issue #147, opened 2026-07-22
Add a log out button


mdeyell-valid-mind:
When we are in auth mode, there should be a logout button that destroys the browsers token

PR #149, opened 2026-07-22
add logout button

mdeyell-valid-mind:
Screenshot 2026-07-22 at 11 59 17
We would like people to be able to log out.

https://github.com/validmind/atryum/issues/147


juanmleng:
Took a walk through this and it's a clean, focused change that does exactly what #147 asked for — nice one! 🙌

The logout button calls `signOut`, which clears the token from local storage via `removeUser()`, so the browser token genuinely gets destroyed rather than just hidden. It only shows when we're in auth mode, and it tucks neatly into the bottom of the sidebar.

One tiny, totally optional thought — not a blocker: `signOut` in `adminAuth.tsx` doesn't wrap `removeUser()` in a try/catch, so if it ever threw you'd get an unhandled rejection from the button's onClick. `refreshAdminAccessToken` already guards the same call as best-effort, so it could be nice to match that someday.

All good from my side!

ui/src/components/Layout.tsx
```diff
@@ -133,6 +138,18 @@ const Layout: React.FC<LayoutProps> = ({ children }) => {
...
           </Stack>
         </VStack>
+        {authStatus === "authenticated" ? (
+          <Box borderTopWidth={1} borderColor="border.base" p={4}>
+            <Button
+              variant="ghost"
+              w="full"
+              justifyContent="flex-start"
+              leftIcon={<Icon as={ArrowRightStartOnRectangleIcon} boxSize={6} />}
+              onClick={() => void signOut()}>
```
hunner, line 148: This signs out of the local atryum instance, but not keycloak (or whatever oauth provider). You can tell because logging out network requests never hit keycloak and when you go to log back in keycloak doesn't prompt for credentials.

It should end up using the `end_session_endpoint` autodiscovered from http://keycloak:8089/realms/atryum/.well-known/openid-configuration I think.

hunner (changes requested):
I think `post_logout_redirect_uri` is a standard url query parameter for oauth logout. [Docs](https://auth0.com/docs/authenticate/login/logout/log-users-out-of-auth0)

juanmleng (approved):

ui/src/components/Layout.tsx
mdeyell-valid-mind: Updated, thanks

hunner (approved):
yassss, works now :)

PR #156, opened 2026-07-24
feat: agent tags (free-form, FOSS primitive for commercial Global Charters)

even-steven:
## Summary
Adds free-form string tags to Atryum agents — a generic, FOSS-usable primitive. Commercial "Global Charters" (organization-wide charters matched to agents by tag) live entirely in the separate ValidMind backend and stay out of the Atryum FOSS UI; Atryum's role is only to store/edit tags and forward them in the evaluate payload so the backend can match against them.

Companion PRs: backend (Global Charters CRUD + matching) + frontend (management UI).

## Changes
- `pkg/migrations/029_agent_tags.go` — adds `agents.tags` (JSON array, default `[]`).
- `internal/store/agents.go` — `AgentRecord.Tags`, encode/decode helpers; the sync `Upsert`'s `ON CONFLICT DO UPDATE` deliberately excludes `tags` so a periodic VM sync never wipes Atryum-native tags.
- `internal/api/handlers.go` — `AdminAgent.Tags`, `AdminAgentInput.Tags` (editable for all agents, including synced ones).
- `internal/invocation/service.go` + `pkg/atryum/atryum.go` + `internal/backend/client.go` — `AgentTags` forwarded through to the backend's `/evaluate` call.
- `ui/src/pages/Agents.tsx` — tag editor (`CreatableSelect`) in the admin UI.

## Testing
- `go build ./...`, `go vet ./...`, `go test ./...` all pass.
- New tests: `internal/store/agents_test.go` (tags survive a sync re-upsert), `internal/api/handlers_test.go` (the local-vs-synced charter-preview branching correctly uses `VMOrganizationCUID`, not `VMCUID`, for manually-created agents).

## Notes for reviewers
- No changes to the FOSS evaluation/rule-matching logic beyond forwarding tags — tags are purely descriptive metadata from Atryum's perspective.


nibalizer (approved):

nibalizer:
Works on my machine

internal/backend/client.go
mdeyell-valid-mind, line 436: should this send tags?

even-steven: Good question — intentionally no. This call hits the backend's hierarchy-only `/api/atryum/unstable/charter-preview` route (backed by `assemble_charter_hierarchy`, not `assemble_effective_charter`), which is what Atryum's own FOSS UI uses to preview an agent's charter chain. Global Charters are commercial-only and have no presence in the Atryum UI, so this endpoint deliberately never matches them — sending tags here wouldn't do anything server-side anyway, since that route doesn't accept a `tags` param.

The ValidMind (commercial) frontend has its own separate preview route (`/agent-authority/agents/{cuid}/charter-preview` in the backend repo) that *does* accept tags and layers in matching global charters — that one's only ever called from the VM frontend, never from Atryum.

Added a comment on `CharterPreviewRequest` (0951b7c) spelling this out so the next person doesn't have to ask the same question.

mdeyell-valid-mind (approved):

PR #137, opened 2026-07-14
scafolding

mdeyell-valid-mind:
> Atryum can now be embedded as a Go dependency by downstream programs. Import github.com/validmind/atryum/pkg/atryum and call `atryum.Main(...)` from the downstream binary’s main package.
  >
  > The extension surface supports:
  >
  > - `atryum.WithRoutes(...)` to register additional HTTP routes on Atryum’s mux.
  > - `atryum.WithMigrations(...)` to run downstream-owned, namespaced schema migrations after Atryum’s built-in migrations.
  > - `atryum.WithDatabase(...)` to receive the initialized database connection after migrations.
  >
  > Example:
  > ```go
  > package main
  >
  > import "github.com/validmind/atryum/pkg/atryum"
  >
  > func main() {
  >     atryum.Main(
  >         atryum.WithRoutes(registerRoutes),
  >         atryum.WithMigrations("my-extension", migrations),
  >         atryum.WithDatabase(bindStores),
  >     )
  > }
  > ```
  > This produces one binary containing the complete Atryum CLI and server plus the downstream extension. Routes contributed through WithRoutes are mounted outside Atryum’s built-in authentication middleware, so downstream handlers are responsible for any authentication they require.

Testing:
I ran atryum and didn't see any changes

PR #157, opened 2026-07-24
Add v0.3.0 changelog; fix stale module path in release ldflags

hunner:
Changelog: agent plan preapproval, embeddable-library extension points (pkg/atryum), server-side session get-or-create, rules-polling for agents, charter preview, logout button, MCP endpoint copy button, Docker Hub image publishing, plus the ClearSessions zombie-watcher fix, two rule-loading fail-safe fixes, and the RecordExecution approval- bypass security fix.

Also fixes release-build's version ldflags: the module path changed from `atryum` to `github.com/validmind/atryum` (#137) but the -X target was never updated, so a release binary would silently report "dev" instead of the tag (confirmed: go build accepts an unmatched -X target without warning). The release-build version self-report check would have caught this at release time, but it's better fixed now than relied on.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,71 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [0.3.0] - 2026-07-24
+
+### Added
+
+- Atryum can now be embedded as a Go library by downstream programs: import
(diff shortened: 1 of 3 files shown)


mdeyell-valid-mind:
`internal/version/version.go` still shows the old `-X`

mdeyell-valid-mind (approved):

mdeyell-valid-mind (approved):

nibalizer (approved):

PR #158, opened 2026-07-24
fix: persist matched_rule_id for AI-decided invocations

hunner:
The proxy Invoke path assigned inv.MatchedRuleID inside a branch covering only DispositionHuman, DispositionWorkflow, and dispositionAIEscalated, so AI-decided invocations — hard denials and auto-approvals — were stored with matched_rule_id = null. The audit view then had no id to look up and reported the still-present rule as "*Deleted Rule*".

Assign MatchedRuleID unconditionally before the disposition switch. The downstream denyByPolicy/executeNow calls carry inv onward and their UpdateResult already writes the column, so the remaining conditional write does only what it must: persist human-gated invocations before the wait so approve/deny handlers can reference the rule. This matches the Submit path, which has always set the id unconditionally.

On the UI side, stop inferring deletion from a missing rule. matched_rule_id is a FK with ON DELETE SET NULL, so a present id can never refer to a deleted rule — it is simply absent from the loaded rules list, and now renders as "*Unknown rule (<id>)*". A null id on an AI decision has two indistinguishable causes, a pre-fix record or a genuine deletion, so it renders as "*Unrecorded or deleted rule*" rather than asserting either.

TestInvokeUsesAuthenticatedAgentIDForRulesAndEvents needed an approval_rules row added: it feeds a rule id from a stub rules store, and the auto-approve path now actually writes that id, which the FK rejects without the row.

The equivalent copy of invocationDisplay.ts in the validmind/frontend repo still carries the old label and needs the same change.

### DIFF
--- a/internal/invocation/agent_id_test.go
+++ b/internal/invocation/agent_id_test.go
@@ -48,6 +48,17 @@ func TestInvokeUsesAuthenticatedAgentIDForRulesAndEvents(t *testing.T) {
 		ServerPatterns: []string{"*"}, ToolPatterns: []string{"*"},
 		Enabled: true,
 	}}}
+	// The stub decides which rule matches, but the id it hands back is still
+	// written to invocations.matched_rule_id, which carries a FK to
+	// approval_rules(id). Auto-approved invocations now persist that id, so the
+	// row has to exist for the write to succeed.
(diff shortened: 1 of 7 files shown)


mdeyell-valid-mind (approved):
i didn't run this but the service.go change looks good

PR #160, opened 2026-07-27
Justfile Docker, k8s example, opencode plugin

nibalizer:
This adds a few nice-to-haves:

- Justfile can now build a docker image from local code
- Example deployment on kubernetes or openshift
- (long overdue) OpenShift harness plugin

### DIFF
--- a/examples/opencode-plugin/atryum.ts
+++ b/examples/opencode-plugin/atryum.ts
@@ -394,6 +394,7 @@ async function submit(
     input: jsonSafe(input),
     request_id: callID,
     thread_id: sessionID || undefined,
+    client_session_id: sessionID || undefined,
     agent_id: AGENT_ID || undefined,
     client_name: CLIENT_NAME,
     client_version: CLIENT_VERSION || undefined,
--- a/examples/opencode-plugin/atryum.ts
+++ b/examples/opencode-plugin/atryum.ts
@@ -1,0 +1,595 @@
+import { exec } from "node:child_process"
+import { createHash } from "node:crypto"
+import { readFileSync } from "node:fs"
+import { mkdir, readFile, rename, writeFile } from "node:fs/promises"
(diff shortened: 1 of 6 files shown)


examples/opencode-plugin/atryum.ts
```diff
@@ -0,0 +1,595 @@
...
+  input: ToolInput,
+  sessionID: string,
+): Promise<InvocationResponse> {
+  const body = {
+    source: SOURCE,
+    tool,
+    description: describe(input),
+    input: jsonSafe(input),
+    request_id: callID,
+    thread_id: sessionID || undefined,
```
mdeyell-valid-mind, line 396: we need to add `client_session_id`

examples/opencode-plugin/commands/atryum.md
```diff
@@ -0,0 +1,24 @@
...
+Set up Atryum approval gating for this OpenCode project.
+
+Reference files:
+- @examples/opencode-plugin/atryum.ts
+- @examples/opencode-plugin/README.md
+
+Workflow:
+1. Briefly explain what the plugin does.
+2. If the user has not already told you their Atryum base URL, ask for it and default to `http://localhost:8080`.
+3. If the user has not already given you a bearer token or token-minting command, explain how to get one from their Atryum web login and ask them to paste it. Wait for the user's reply before making file changes.
```
mdeyell-valid-mind, line 14: should this also consider the existence of no auth mode?

nibalizer: I sorta don't think this is even a good idea and I will clean it up later.

examples/opencode-plugin/atryum.ts
nibalizer: added

mdeyell-valid-mind (approved):
i don't have open code to test with but looks good

PR #164, opened 2026-07-28
Use underscore names for synthetic plan MCP tools

nibalizer:
## Summary
- rename Atryum's synthetic MCP plan helpers from dotted names to `atryum_plan_submit` and `atryum_plan_get`
- derive agent-facing initialize/tool descriptions from the shared helper-name constants so the advertised names cannot drift
- add a regression test that fails if `tools/list` ever returns dotted synthetic tool names again
- update related auth/plan tests and docs/examples to use the underscore names

Also includes a commit where I cleaned up messaging in some comments after discussing auth with amp.

## Why
Anthropic/Claude rejects dotted tool names during tool registration, so the previous synthetic helper names (`atryum.plan.submit` / `atryum.plan.get`) could break MCP tool exposure even though they were otherwise valid MCP identifiers.

## Testing
- `git diff --check -- README.md examples/codex-mcp/README.md internal/api/auth_test.go internal/api/handlers.go internal/api/handlers_test.go internal/invocation/local_evaluator.go internal/invocation/plan_service.go internal/invocation/plan_service_test.go`
- not run: `go test` / `gofmt` are unavailable in this container (`go` and `gofmt` not on `PATH`)


mdeyell-valid-mind (approved):
makes sense

